PR #5, opened 2021-10-26
Fixed an issue with the calling-user decorator [very small]

neketka:
### Summary 
The calling-user decorator is used to get a user from a socket.io call. It had a small error that needed to be fixed

- [x] Fixed property access

Remaining TODOs:

- [ ] Implement authentication in totality

### Test Plan 

Not applicable


markd88 (approved):
The PR makes a small change and it looks good

PR #8, opened 2021-10-31
add jwt and local strategy

markd88:
### Summary 

This PR adds files for "jwt" and "local" strategies, which are for future Route Gaurds.

Remaining TODOs:

- 1. figure out ways to extract id from different authentication types
- 2. get a look at the database structure

### Test Plan 
1. Postman to test the apis


server/src/auth/strategy/jwt.strategy.ts
```diff
@@ -0,0 +1,19 @@
+import { Injectable } from '@nestjs/common';
+import { PassportStrategy } from '@nestjs/passport';
+import { ExtractJwt, Strategy } from 'passport-jwt';
+// import { jwtConstants } from '../constants';
+
+@Injectable()
+export class JwtStrategy extends PassportStrategy(Strategy) {
+  constructor() {
+    super({
+      jwtFromRequest: ExtractJwt.fromAuthHeaderAsBearerToken(),
```
neketka, line 10: We don't actually use this because WebSockets don't have these sorts of headers. We use them as a part of the body.

server/src/auth/strategy/local.strategy.ts
```diff
@@ -0,0 +1,19 @@
...
+import { Strategy } from 'passport-local';
+import { AuthService } from '../auth.service';
+
+@Injectable()
+export class LocalStrategy extends PassportStrategy(Strategy) {
+  constructor(private readonly authService: AuthService) {
+    super();
+  }
+
+  async validate(username: string, password: string): Promise<any> {
```
neketka, line 12: We actually do have a type for users. It's either Admin or User. Be careful using "any" since it will subvert type-checking. Also, we don't actually use passwords for authentication. We use tokens. AuthService should have something on that.

neketka (changes requested):
I left a few changes. Once we discuss authentication, I think you'll understand where to head next. You are allowed to have a PR over multiple sprints.

server/src/auth/auth.service.ts
```diff
@@ -16,14 +22,68 @@ export class AuthService {
 
   /** Get identifier from a google token */
   getIdFromGoogleToken(token: string): string | undefined {
-    return undefined;
+
+    return undefined
+  }
+
+  
+  async verifyGoogle(token_id : string): Promise<any> {
```
neketka, line 30: Don't use "any" as a type when more specific options are available.

server/src/auth/auth.service.ts
```diff
@@ -16,14 +22,68 @@ export class AuthService {
...
+    try {
```
neketka, line 31: The stuff in this method should be in getIdFromGoogleToken. It returns undefined if no id could be found/google/verifyIdToken fails, or it returns the id (that's all we need).

server/src/auth/auth.service.ts
```diff
@@ -16,14 +22,68 @@ export class AuthService {
...
+      if (error) {
+        return "error"
+      }
+    }
+    return "error"
   }
 
+
+  async loginGoogle(ID_token : string) {
+    const payload = await this.verifyGoogle(ID_token);
```
neketka, line 51: You can factor out the stuff in this method to use more than just Google. If you skip the verifyGoogle part and have the method take in AuthType, you can rename this method to "login" so you won't have to repeat the code.

server/src/auth/google/google.controller.ts
```diff
@@ -1,4 +1,14 @@
...
 
 @Controller('google')
-export class GoogleController {}
+export class GoogleController {
+  constructor(private readonly authService: AuthService) {}
+
+  // login
+  @Post('/auth/login')
+  async login(@Request() req) {
+    return this.authService.loginGoogle(req.token);
```
neketka, line 11: If you make the change I mentioned above, you can call it "login(req.body.token, AuthType.GOOGLE)." Also, please use "@Body()" instead of "@Request()" and change "@Post('...')" to "@Post()" because "@Controller('google')" already defines this route as "/google" for this controller.

neketka (changes requested):
Good job on adding the Google authentication. This will move us very far along getting "login" to work. Just left a few minor comments that will make this code even more helpful. Also, do "npm run format" and "npm run build" before committing to this pr.

lobsterdude3 (commented):
Looks good to me, I opened the tutorial and it seems similar to it.

neketka (approved):

PR #10, opened 2021-12-17
Add support for clickFunc and a support for editable props

lobsterdude3:
### Summary 

- [x] implemented buttons that appear when a clickFunc is with the string
- [x] made support for editable props

Remaining TODOs:

- [ ] figure out the best way to call the edit function
- [ ] figure out what the edit function will be used for

]

### Test Plan 

Create sample data that has click functions, see if it works


markd88 (commented):
The code made some changes about the layout and it looks good

neketka (approved):

PR #11, opened 2021-12-19
Added state interfaces

neketka:
### Summary 

Adds definitions for front-end state management.

- [x] Rewrote backend model in the frontend

Remaining TODOs:

- [ ] Implement react recoil system

### Test Plan 

Check that front-end state is the same as the backend model.

### DIFF
--- a/admin/src/state/challenge.ts
+++ b/admin/src/state/challenge.ts
@@ -1,0 +1,32 @@
+
+import { EventBase } from './event-base';
+import { PrevChallenge } from './prev-challenge';
+
+/**
+ * Entity describing a challenge associated with a place
+ */
+export interface Challenge {
+  id: string;
+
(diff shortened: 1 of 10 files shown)


admin/src/state/challenge.ts
```diff
@@ -0,0 +1,32 @@
...
+import { PrevChallenge } from './prev-challenge';
+
+/**
+ * Entity describing a challenge associated with a place
+ */
+export interface Challenge {
+  id: string;
+
+  /** Index of the challenge relative to others in the linked event, -1 if last */
+  eventIndex: number;
```
lobsterdude3, line 12: Spacing should be consistent with all the types, otherwise the code looks fine

lobsterdude3 (commented):
Needs to fix spacing, also some types do not have comments

PR #24, opened 2022-02-17
Fix android pixel build error

lobsterdude3:
### Summary 

Fixed a build error that only I had. Works on running on pixel XL 29

### Test Plan 
 Someone else test and see if it runs


neketka (approved):
Tested it and it seems to still work. It does install Android 31 though.

PR #30, opened 2022-03-03
Added setCurrentEvent to the group gateway and createEventTracker to the event service

youssef-attia:
### Summary 

- [x] implemented setCurrentEvent and createEventTracker

### Blockers 

### Breaking Changes  


cathli66 (approved):
Looks great! Maybe there is some way to reduce code reuse in the try catch block?

neketka (changes requested):
Please resolve the CI issues before merging.

server/src/group/group.gateway.ts
```diff
@@ -24,6 +25,7 @@ export class GroupGateway {
   constructor(
     private clientService: ClientService,
     private groupService: GroupService,
+    private eventService: EventService,
```
neketka, line 28: You added EventService here, but GroupModule does not import EventModule in its 'imports' array. This code will not start if you were to run 'docker compose up'.

neketka (changes requested):

server/src/group/group.gateway.ts
```diff
@@ -76,5 +76,32 @@ export class GroupGateway {
...
+    const userGroup = await userMember.group.load();
+    let newEvent = (await this.eventService.getEventsByIds([data.eventId]))[0];
+    (await userGroup.members.loadItems()).forEach(
+      async (member: GroupMember) => {
+        //if the user already has the event, keep their tracker
+        let currentUser = await member.user.load();
+        try {
+          let memberEvent =
+            await this.eventService.getCurrentEventTrackerForUser(currentUser);
+          if (memberEvent.event.id != data.eventId)
```
neketka, line 90: I don't think this condition is necessary

server/src/group/group.gateway.ts
```diff
@@ -76,5 +76,32 @@ export class GroupGateway {
...
+      },
+    );
+    userGroup.currentEvent.set(newEvent);
+    await this.groupService.saveGroup(userGroup);
+    let updateGroupData: UpdateGroupDataDto = {
+      curEventId: data.eventId,
+      members: [],
+      removeListedMembers: false,
+    };
+    this.clientService.emitUpdateGroupData(user, updateGroupData);
```
neketka, line 105: Notify all members, not just this one.

neketka (changes requested):
Last few changes, but afterwards this should be ready to go!

neketka (approved):
LGTM

PR #31, opened 2022-03-03
Added requestGroupData in GroupGateway and getGroupForUser in GroupService

cathli66:
### Summary 

This pull request implements part of GroupGateway.

- [x] implemented requestGroupData
- [x] implemented getGroupForUser

Remaining TODOs:

- [ ] implement joinGroup
- [ ] implement leaveGroup

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,8 @@
+{
+  "dependencies": {
+    "@mikro-orm/core": "^5.0.5",
+    "@mikro-orm/nestjs": "^5.0.0",
+    "@mikro-orm/postgresql": "^5.0.5",
+    "mikro-orm": "^5.0.5"
+  }
+}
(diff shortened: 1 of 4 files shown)


youssef-attia (commented):
Everything you've added makes sense to me. I like that you made a method in Group Service to return the group info. I'll probably have to update my method so that I'm using that. I was curious whether it was relevant for when you're setting the curChallengeId on line 49 if it matters that getCurrentEventTrackerForUser might fail? I was told it would be better to put those kinds of issues in try catch blocks just in case but is there something in your code that makes this not necessary? Thanks

neketka:
>I was curious whether it was relevant for when you're setting the curChallengeId on line 49 if it matters that getCurrentEventTrackerForUser might fail? 

Assuming that both joinGroup and setCurrentEvent preserve the invariant that a user has an event tracker for every event they've been in, this shouldn't ever fail.

neketka (changes requested):
Make sure to fix the event id and code style issue, then it should be good to go.

neketka (approved):
LGTM

PR #32, opened 2022-03-03
Added state management for flutter app.

neketka:
### Summary

This pull request is the first step towards implementing state management for flutter.

- [x] Created models
- [x] Add dependency injection for the models
- [x] Fixed up API  

Remaining TODOs:

- [ ] Switch from DTOs to model classes

### Test Plan 

Test if the model works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,10 @@ npm start
 
 ```
 Open the main.dart file, then click the Play button
+*OR* to run with a custom server url
+
+cd game
+flutter run --dart-define=API_URL=https://example.com
 ```
 
 ## Contributors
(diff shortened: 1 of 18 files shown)


nirbhaysinghnarang (approved):
__Before I could review this PR, I had to learn more about Flutter state management using the _provider_ package. For anyone else going through this code, consider using the following resources:

- https://docs.flutter.dev/development/data-and-backend/state-mgmt/simple
- https://pub.dev/packages/provider

With that said, I also feel that a little bit of documentation (perhaps only for each class like the GameServer and APIClient classes) could have helped me better understand the code.

Also, for anyone else reviewing this PR:

- run _pod install_ in the _ios_ folder 
- run _flutter pub get_ (VSCode should probably do this automatically)

PR #38, opened 2022-03-08
Implemented group.gateway.joinGroup, leaveGroup.

cathli66:
### Summary

This pull request implements part of GroupGateway.

- [x] implemented requestGroupData
- [x] implemented joinGroup
- [x] implemented leaveGroup


server/src/group/group.service.ts
```diff
@@ -53,4 +54,46 @@ export class GroupService {
     const groupMember = await user.groupMember!.load();
     return groupMember!.group.load();
   }
+
+  /** Get group from the id. 
+   * Throws a NotFoundError if the id does not correspond to a group. */
+  async getGroupFromID(id: string): Promise<Group> {
+    const group = await this.groupsRepository.findOne({ friendlyId: id })
+    if (group == null)
+      throw NotFoundError
```
cathli66, line 63: I'm not sure if throwing an error is the best here, or if there is a better way to handle possible invalid ids. Could I just assume that all id's are correct as a precondition?

server/src/group/group.gateway.ts
```diff
@@ -64,13 +64,19 @@ export class GroupGateway {
   async joinGroup(
     @CallingUser() user: User,
     @MessageBody() data: JoinGroupDto,
-  ) {}
+  ) {
+    this.groupService.joinGroup(user, data.groupId);
+    this.requestGroupData(user, {});
```
neketka, line 69: Update both members of the old group and new group, not just the one person.

server/src/group/group.gateway.ts
```diff
@@ -64,13 +64,19 @@ export class GroupGateway {
...
+  }
 
   @SubscribeMessage('leaveGroup')
   async leaveGroup(
     @CallingUser() user: User,
     @MessageBody() data: LeaveGroupDto,
-  ) {}
+  ) {
+    this.groupService.leaveGroup(user);
+    this.requestGroupData(user, {});
```
neketka, line 78: Update both members of the old group and new group, not just the one person.

server/src/group/group.gateway.ts
```diff
@@ -64,13 +64,19 @@ export class GroupGateway {
   async joinGroup(
     @CallingUser() user: User,
     @MessageBody() data: JoinGroupDto,
-  ) {}
+  ) {
+    this.groupService.joinGroup(user, data.groupId);
```
neketka, line 68: await on this

server/src/group/group.gateway.ts
```diff
@@ -64,13 +64,19 @@ export class GroupGateway {
...
+    this.requestGroupData(user, {});
+  }
 
   @SubscribeMessage('leaveGroup')
   async leaveGroup(
     @CallingUser() user: User,
     @MessageBody() data: LeaveGroupDto,
-  ) {}
+  ) {
+    this.groupService.leaveGroup(user);
```
neketka, line 77: await on this

server/src/group/group.service.ts
```diff
@@ -53,4 +54,47 @@ export class GroupService {
     const groupMember = await user.groupMember!.load();
     return groupMember!.group.load();
   }
+
+  /** Get group from the id.
+   * Throws a NotFoundError if the id does not correspond to a group. */
+  async getGroupFromID(id: string): Promise<Group> {
+    const group = await this.groupsRepository.findOne({ friendlyId: id });
+    if (group == null) throw NotFoundError;
```
neketka, line 62: It would throw an exception if you do "findOneOrFail"

server/src/group/group.service.ts
```diff
@@ -53,4 +54,47 @@ export class GroupService {
     const groupMember = await user.groupMember!.load();
     return groupMember!.group.load();
   }
+
+  /** Get group from the id.
+   * Throws a NotFoundError if the id does not correspond to a group. */
+  async getGroupFromID(id: string): Promise<Group> {
```
neketka, line 60: Maybe call this "byFriendlyId", so it's not confused with the real id.

server/src/group/group.service.ts
```diff
@@ -53,4 +54,47 @@ export class GroupService {
...
+    await this.groupsRepository.persistAndFlush(group);
+  }
+
+  /** Moves the user out of their current group into a new group */
+  async leaveGroup(user: User) {
+    const oldGroup = await this.getGroupForUser(user);
+    const newGroupId = (
+      await this.createFromEvent(await oldGroup.currentEvent.load(), user)
+    ).friendlyId;
+    this.joinGroup(user, newGroupId);
```
neketka, line 98: await on this

server/src/group/group.service.ts
```diff
@@ -53,4 +54,47 @@ export class GroupService {
...
+  async removeUserFromGroup(user: User) {
+    const oldGroup = await this.getGroupForUser(user);
+    oldGroup.members.remove(await user.groupMember!.load());
+    if (oldGroup.members.length == 0)
+      this.groupsRepository.removeAndFlush(oldGroup);
+  }
+
+  /** Adds user to an existing group, given by the group's id. */
+  async joinGroup(user: User, joinId: string) {
+    this.removeUserFromGroup(user);
```
neketka, line 77: await on this

server/src/group/group.service.ts
neketka, line 98: createFromEvent already joins the person into a new group as far as I'm aware

server/src/group/group.service.ts
```diff
@@ -53,4 +54,47 @@ export class GroupService {
...
+   * Throws a NotFoundError if the id does not correspond to a group. */
+  async getGroupFromID(id: string): Promise<Group> {
+    const group = await this.groupsRepository.findOne({ friendlyId: id });
+    if (group == null) throw NotFoundError;
+    return group;
+  }
+
+  /** Helper function that removes a user from their current group.
+   * If the user is the only member, deletes the group. */
+  async removeUserFromGroup(user: User) {
```
neketka, line 68: Consider returning Promise<Group | null> where it returns group if it still exists and null if it does not. It'll help you when you are making members of the old group (if the group exists that is) aware that someone has left.

server/src/group/group.service.ts
```diff
@@ -53,4 +54,47 @@ export class GroupService {
...
+    const oldGroup = await this.getGroupForUser(user);
+    oldGroup.members.remove(await user.groupMember!.load());
+    if (oldGroup.members.length == 0)
+      this.groupsRepository.removeAndFlush(oldGroup);
+  }
+
+  /** Adds user to an existing group, given by the group's id. */
+  async joinGroup(user: User, joinId: string) {
```
neketka, line 76: same thing as in removeUserFromGroup, but return removeUserFromGroup's value at the end of the function

server/src/group/group.service.ts
```diff
@@ -53,4 +54,47 @@ export class GroupService {
...
+
+    group.members.set([groupMember]);
+    user.groupMember?.set(groupMember);
+    groupMember.user.set(user);
+
+    await this.groupsRepository.persistAndFlush(group);
+  }
+
+  /** Moves the user out of their current group into a new group */
+  async leaveGroup(user: User) {
```
neketka, line 93: Same thing as in removeUserFromGroup, but you'll have to do this some other way than joinGroup since you do not use removeUserFromGroup.

neketka (changes requested):
Left some changes for you to make. Hopefully, this clarifies things!

neketka (approved):
LGTM

PR #41, opened 2022-03-11
Testing for Google Log In.

nirbhaysinghnarang:
### Summary 

This pull request is the first step towards completion of Authentication.

- [ x] implemented Sign In with Google.
- [ x] Implemented auto sign in if the user is already signed in.
- [ x] created links to home page from sign in page.

TODOS:
- change state of app based on user data
- add screen to change username
- add photo automatically based on name attributes.


cathli66 (commented):
Great work! This helped me understand how flutter works with the backend.

youssef-attia (commented):
I feel like a lot of the database connection is really specific so I don't really understand all the work with checksums and stuff. I did get an understanding of the strategy used to make sure the user has their location turned on for the app. I also found it interesting that I'm seeing some of the data properties that come up in the database being sent during the login response. The structure of a widget in Flutter really follows an object model I feel and it seems to make sense to me. Great work overall.

game/lib/login/login_page.dart
```diff
@@ -0,0 +1,123 @@
...
+              ],
+            ),
+            Column(
+              children: [
+                Padding(
+                  padding: const EdgeInsets.only(bottom: 30),
+                  child: SignInButton(
+                    Buttons.Google,
+                    onPressed: () async {
+                      final client = ApiClient(widget.storage, widget.API_URL);
```
neketka, line 79: You should not be creating a new ApiClient, but consuming the one at the top of the hierarchy. Use the consumer widget for that.

game/lib/login/login_page.dart
```diff
@@ -0,0 +1,123 @@
...
+  void initState() {
+    super.initState();
+    WidgetsBinding.instance?.addPostFrameCallback((_) {
+      //to auto log in a user if already logged in.
+      checkLoggedIn();
+    });
+  }
+
+  Future<void> checkLoggedIn() async {
+    if (await ApiClient(widget.storage, widget.API_URL).tryRelog()) {
```
neketka, line 31: Same here, don't make a new one.

neketka (changes requested):
Make sure to use the correct ApiClient when you're writing your code.

neketka (approved):
Looks complete!

PR #44, opened 2022-03-20
Chris

cgu2020:
### Summary 

This pull request is the first step towards implementing checkForRewards and the completedChallenge gateway.

- [ ] Checks if the user has completed the event
- [ ] Creates a new reward for finished perpetual events
- [ ] Checks if the user finishes in time for timed-events
- [ ] Finds a reward and allocates to user
- [ ] Updates the client


server/src/challenge/challenge.service.ts
```diff
@@ -2,22 +2,29 @@ import { InjectRepository } from '@mikro-orm/nestjs';
...
 import { User } from '../model/user.entity';
+import { EventTracker } from 'src/model/event-tracker.entity';
 
 @Injectable()
 export class ChallengeService {
   constructor(
     private userService: UserService,
     @Inject(forwardRef(() => EventService))
     private eventService: EventService,
+    @Inject(forwardRef(() => EventService))
```
neketka, line 20: You don't need this inject here.

server/src/challenge/challenge.service.ts
```diff
@@ -2,22 +2,29 @@ import { InjectRepository } from '@mikro-orm/nestjs';
...
+    private clientService: ClientService,
     @InjectRepository(Challenge)
     private challengeRepository: EntityRepository<Challenge>,
     @InjectRepository(PrevChallenge)
     private prevChallengeRepository: EntityRepository<PrevChallenge>,
+    @InjectRepository(PrevChallenge)
```
neketka, line 26: Switch from PrevChallenge to EventReward

server/src/challenge/challenge.service.ts
```diff
@@ -132,4 +139,39 @@ export class ChallengeService {
...
+      return userRewards
+    }
+    
+    if(rewardType === EventRewardType.PERPETUAL){
+      const newReward = await this.rewardRepository.findOne({event: eventBase});
+      if(newReward !== null){
+        newReward.claimingUser = user
+        newReward.isRedeemed = false
+        await this.rewardRepository.persistAndFlush(newReward);
+        this.clientService.emitUpdateRewardData(user, newReward)
```
neketka, line 163: Make sure to use the DTO when sending data to the user (you would not be sending newReward to the user).

server/src/challenge/challenge.service.ts
```diff
@@ -132,4 +139,39 @@ export class ChallengeService {
...
+    }
+
+    const rewardType = eventBase.rewardType
+    if(rewardType === EventRewardType.LIMITED_TIME_EVENT && eventBase.time > new Date()){
+      return userRewards
+    }
+    
+    if(rewardType === EventRewardType.PERPETUAL){
+      const newReward = await this.rewardRepository.findOne({event: eventBase});
+      if(newReward !== null){
```
neketka, line 159: You're gonna want to duplicate newReward here. this.rewardRepository.create({ ...newReward, id: v4() }) should make you a copy

server/src/challenge/challenge.service.ts
neketka, line 163: Also, you should probably not be using clientService inside of a service, use it in a gateway.

server/src/challenge/challenge.service.ts
```diff
@@ -132,4 +139,39 @@ export class ChallengeService {
 
     return eventTracker;
   }
+
+  /** Check if the current event can return rewards */
+  async checkForReward(user: User, eventBase: EventBase, eventTracker: EventTracker) {
```
neketka, line 144: You might actually want to return Promise<EventReward | null> from this function and return a reward or null depending on what you do in this function. That way you can use the reward you picked out in this function (if it's not null) and call emitUpdateUserData from the gateway.

neketka (changes requested):
Take a look at the comments I posted.

PR #53, opened 2022-04-11
Add deployment.

neketka:
### Summary 

Add deployment files and fix bugs with the flutter app.


nirbhaysinghnarang (approved):

nirbhaysinghnarang (commented):
For the progress bar for distance to an event, I feel that having the transparent progress indicator move from left to right (grows as user gets closer) makes more intuitive sense (to me at least) than moving it from right to left (grows smaller as user gets closer). 
Besides that, in the next update to the mobile UI, I think we should work on increasing the responsiveness of the design because I think we have a lot of 'magic numbers' in the frontend codebase rn. For example, line 432 on _HomePageWidgetState. These 'magic numbers' might make the UI look weird on certain viewports.

PR #55, opened 2022-04-15
Added profanity filter

nirbhaysinghnarang:
Added profanity filter to Flutter frontend.


cathli66 (commented):
Although my Flutter knowledge is limited, this looks fine by me. Does the profanity filter package used take into account words that are not standalone, like "f***hi"? If not, how are you planning to implement a check for these situations?

youssef-attia:
From what I see, you are pulling an array of profanities from some API and checking the usernames to see if any of the usernames contain a profanity. I'm wondering if some limit on the ascii characters that users are able to enter for their username. I feel like if not, this may add several possibilities of bypassing the filter through alternate ascii representations but at the same time we don't want to limit user freedom too much. Otherwise, the logic of the code looks great to me. Good work.

cgu2020 (commented):
Seems good. However, is it possible to just include the profanity list into the app rather than fetch the data from the url? I am not a frontend developer, but I feel like it would be easier and faster to include the txt file if possible.

game/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -525,7 +545,11 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
-				CURRENT_PROJECT_VERSION = "$(FLUTTER_BUILD_NUMBER)";
+				CODE_SIGN_ENTITLEMENTS = Runner/Runner.entitlements;
+				CODE_SIGN_IDENTITY = "Apple Development";
+				CODE_SIGN_STYLE = Automatic;
+				CURRENT_PROJECT_VERSION = 3;
+				DEVELOPMENT_TEAM = U9S73WTH2W;
```
neketka, line 552: Lots of weird config things that should probably not be in this PR, so I don't want to approve it just yet until we know that this won't affect our iOS build.

neketka (changes requested):
Check comment.

PR #57, opened 2022-04-21
Added restrictions.

neketka:
### Summary

This pull request is the first step toward implementing restrictions on users.

- [x] Created RestrictionGroup entity with relationships with users and events
- [x] Added restricted user generation for admin dashboard and restriction management
- [x] Added DTOs and client-side API

Remaining TODOs:

- [ ] Add UI for admin frontend to use this feature.

### Test Plan 

Use Postman to test the API + implement frontend and test


erichuang27 (approved):
Looked through all the code, seems to cover most of the connections with the restriction group. I think further testing would be needed with the front end and check the leaderboard and groups and such but it shouldn't break anything.

PR #58, opened 2022-04-22
Global Leaderboard and Rewards Cell

BrianLa0616:
### Summary 

Created Global leaderboard and changed text size of Rewards Cell to be larger

- [x] implemented Global Leaderboard
- [x] Rewards cell

Remaining TODOs:

- [ ] Test rewards page


neketka (changes requested):
image
Something is not going right here. Also, the global leaderboard is not showing anything. Maybe name the old leaderboard to "Event Leaderboard" as well?

neketka (approved):
LGTM

youssef-attia (commented):
I like the hierarchy of GlobalLeaderboardWidget to leaderBoardUserCell and leaderBoardCell. What I found really interesting is your data management with the parameters you passed through. Something I did not know about before was on line 34 of the global_leaderboard_widget file, where you use a builder with the contexts to pass the information to the elements below. I looked it up and it seems a lot easier than making a standalone widget first. Good work!

PR #59, opened 2022-04-27
Rewarded page

youssef-attia:
### Summary 

Added a rewarded page that pops up when a reward is completed. 
It takes the game state as a parameter and uses it to display the challenge name in the text and the challenge image as a blurred background.


neketka (approved):
Works great! LGTM

PR #60, opened 2022-04-27
Restriction tab on the admin frontend

cgu2020:
### Summary 
Restriction tab can create, edit, and delete restrictions.

- [x] Create restriction
- [x] Edit restriction
- [x] Delete restriction
- [x] Add event
- [x] Clear events


neketka (approved):
I fixed some issues on the backend (that I seemingly missed when making my PR) related to cascading. Otherwise, LGTM!

server/src/model/user.entity.ts
```diff
@@ -84,11 +85,15 @@ export class User {
   group!: IdentifiedReference<Group>;
 
   /** Event trackers for each event the player participated in */
-  @OneToMany(() => EventTracker, ev => ev.user)
+  @OneToMany(() => EventTracker, ev => ev.user, {
+    cascade: [Cascade.SCHEDULE_ORPHAN_REMOVAL, Cascade.REMOVE],
```
BrianLa0616, line 89: I really like how the property cascade was taken advantage of. I didn't realize that there was this function that allows you to remove elements from the backend so easily. I like how instead of trying to remember all references to the deleted element, you can implement cascade and essentially have the database do the work for you. Very nice!

PR #63, opened 2022-05-03
Implemented database error correction every 15 minutes

cli66:
### Summary 

- [x] Implemented admin.service.checkEventTrackers
- [x] Implemented admin.gateway.checkErrors

### DIFF
--- a/server/src/admin/admin.service.ts
+++ b/server/src/admin/admin.service.ts
@@ -411,6 +411,9 @@ export class AdminService {
 
   async checkEventTrackers() {
     const users = await this.userRepository.findAll();
+    var invalidateRewardData = false;
+    var invalidateEventData = false;
+    var invalidateLeaderboardData = false;
     for (const usr of users) {
       const all_trackers = usr.participatingEvents;
       var score = usr.score;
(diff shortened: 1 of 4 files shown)


server/src/admin/admin.service.ts
```diff
@@ -406,4 +408,67 @@ export class AdminService {
...
+          }
+        }
+        if (!dup_found) trackers.push(t);
+      }
+      // updates score if some events have been deleted
+      var new_score = 0;
+      for (const t of usr.participatingEvents) {
+        new_score += t.eventScore;
+      }
+      usr.score = new_score;
```
cathli66, line 461: Is this the right way to set the user's score?

cathli66:
I'm not sure how to fix the missing dependency issue. I installed scheduler on my local repo, but I'm guessing that doesn't push to GitHub.

neketka (changes requested):

cli66:
When running npm run start, I get an error about postgres:
[query] select 1 from pg_database where datname = 'postgres' [took 3 ms]
Is this an issue?

server/src/admin/admin.service.ts
```diff
@@ -403,4 +405,65 @@ export class AdminService {
...
+      // save data in the userRepository
+      await this.userRepository.persistAndFlush(usr);
+      // changing user's score may change leaderboard positions
+      invalidateLeaderboardData = true;
+    }
+    this.clientService.emitInvalidateData({
+      userEventData: invalidateEventData,
+      userRewardData: invalidateRewardData,
+      winnerRewardData: false,
```
cathli66, line 463: Is it possible for this to be invalidated too?

neketka:
Likely irrelevant with the prisma migration

PR #78, opened 2022-10-27
Fixed race conditions surrounding fixOrDeleteGroup

cathli66:
### Summary 

This pull request wraps possible race conditions with adding/deleting from groups in prisma transactions.


youssef-attia (commented):
Great work, I'm not sure if the PR is finished but I am wondering if there was any logic that needed to be changed in fixOrDeleteGroup to fix the race condition? From what I can tell, the changes are related to migrating from microORM to prisma. Was the migration actually enough to fix the issue?

cgu2020 (commented):
Good work! Not sure why you return at the beginning of the services? I'm probably just missing something with prisma. Everything else looks good!

BrianLa0616:
Looks good! I was wondering how you tested your code and whether race conditions were actually solved?

neketka (approved):
LGTM

PR #79, opened 2022-10-28
Alert Modal

cgu2020:
### Summary 

This pull request is the first step towards implementing admin page exception alerts

- [x] Alert modal popup when there's an exception in the internal server or database

Remaining TODOs:

- [ ] Only error for deleting default event is coded. Other specfic errors will be added as we discover them.


cathli66:
The alert modal looks great! Does the issue with the admin page breaking after any error have to do with not correctly dealing with websockets while catching the error?

nirbhaysinghnarang (commented):
Good work with the modal and error handling! I'm trying to learn more about sockets to be able to contribute to our backend and going over your changes helped!

cgu2020:
The todo occurs with master branch as well so it's not an error modal issue but probably error handling

BrianLa0616:
Looks great! It makes it a lot easier to know what's happening in the backend from the admin page with the alerts!

youssef-attia (commented):
Great work! The connection between frontend and backend here is obviously very strict based on the nature of the task so it is great to see you come up with a solution that works by sending an entry in the backend that is then responded to in the front end. I was wondering if we needed all the console.logs for a reason since we don't see those in production as far as i can tell?

PR #80, opened 2022-10-28
Device login.

nirbhaysinghnarang:
PR for device login and restriction generalizing.


server/src/event/event.service.ts
```diff
@@ -44,6 +44,13 @@ export class EventService {
     return true;
   }
 
+  /** Retrieves default event */
+  async getDefaultEvent() {
+    return await this.prisma.eventBase.findUniqueOrThrow({
+      where: { isDefault: true },
+    });
+  }
+
```
cli66, line 50: We have just one default event for now, but in the future the event retrieved should depend on the organization (assuming diff orgs have diff default events?)

cli66: Actually this method isn't used anywhere, should I just delete it

PR #81, opened 2022-10-30
Added support for internal logging of system events

youssef-attia:
### Summary 

A new table has been created to store logs of user events such as joining or leaving a group, logging in, creating deleting or editing users, changes related to events, challenges, and rewards, and more. This will make it easier to debug breakages or misuses in the system once it is published.

### Test Plan 

Run the server and trigger several different actions and then preview the logging table to insure it includes all needed information.


neketka:
No longer necessary, too many conflicts

youssef-attia:
Continued in #102

PR #83, opened 2022-11-14
Enable and Log Events Using Google Analytics

youssef-attia:
### Summary 

Google Analytics is used in apps to get an idea of what services users are relying on and what parts of the app may be causing trouble for them. There are several different types of events that can be logged. Some events are logged automatically as listed here: [Automatically Logged Events](https://support.google.com/firebase/answer/9234069). And some are logged manually. So far, the events we will log manually will be similar to the events logged in #81 but will also include user logins, signups, and log outs and some more internal events.

Remaining TODOs:

- [ ] Enable Google Analytics on the CornellGo Firebase account
- [ ] Inject logs of specified events into their appropriate call locations in the flutter app

### Test Plan 

Basic usage of the app with a focus on triggering specified events that will be logged and checking the analytics page for the appropriate logs.

### Blockers 

Enabling Google Analytics on the CornellGo Firebase account


youssef-attia:
Closed until GAnalytics is enabled by Cornell

PR #84, opened 2022-11-16
Setup GH Workflow for EC2 CI/CD

nirbhaysinghnarang:
Implement CI/CD to AWS EC2 using Github Actions.


neketka:
We will stick to heroku for now

PR #86, opened 2022-12-09
Frontend Documentation

cgu2020:
### Summary 

This pull request is the first step towards implementing documentation

- [x] comments on the frontend admin page

Remaining TODOs:

- [ ] create documentation for react modules


nirbhaysinghnarang (commented):
I think we should all focus on writing more comments in our code and follow specific formatting. For TS, for example, maybe we could follow https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html

BrianLa0616 (commented):
When I was working on the front end, I thought it was difficult to keep track of sockets. These comments would definitely help with that!

PR #87, opened 2022-12-09
Private DockerHub Image

nirbhaysinghnarang:
### Summary 

In this PR, I added support for the Dockerized image being stored as a private image on DockerHub. This included changing permissions on DockerHub, creating a dockerconfig file and then uploading that to S3. This dockerconfig file is read in Dockerrun.aws.json which pulls and reads it from S3 to get authentication credentials and then retrieves the docker image, proceeding as normal.

TODOs:
1. Fix IAM role on AWS for EB. Current Error: `S3 Permissions Denied`.


neketka:
No longer necessary

PR #88, opened 2022-12-09
Flutter Clean Code

youssef-attia:
### Summary 

Cleaned up some indentation in the flutter app for readability.


cli66:
Cleaning up the Flutter formatting will really help with the readability of the frontend, which is especially helpful since we will be doing lots of redesigns next semester. Great work!

PR #93, opened 2023-02-10
Favorite + Filtering Events

cli66:
### Summary 

Keeps track of favorite events that a user marked, filters events given a new/saved/finished filter based on the status of the event's challenges.

- [x] implemented setFavorite in event.service.ts + gateway methods
- [x] updated the user and event DTO
- [x] updated schema and db
- [x] implemented requestFilteredEvents + gateway methods
- [x] added cursor-based pagination
- [x] tested in Postman

Remaining TODOs:

- [ ] implement in flutter (frontend)

### Notes 

- Database schema change (added favorites EventBase[] to Users and userFavorite User[] in EventBase)


erichuang27:
Run prettier and resubmit, also should test in postman as well before merging. Looks good otherwise

erichuang27 (changes requested):
Run prettier. Also you should test in postman as well.

cgu2020 (commented):
We should also add an unfavorite gateway unless updateUserData works as an unset favorite. Otherwise, this looks good!

neketka:
> We should also add an unfavorite gateway unless updateUserData works as an unset favorite. Otherwise, this looks good!

We should not, setFavorite already takes in a bool

server/src/event/event.service.ts
```diff
@@ -443,6 +443,10 @@ export class EventService {
           ? EventRewardType.LIMITED_TIME
           : EventRewardType.PERPETUAL,
       endTime: new Date(event.endTime),
+      // rewards: {set: event.rewardIds.map(id => ({ connect: {id: id} })) },
+      // challengeIds: event.challengeIds,
```
erichuang27, line 447: do you still need this stuff or can it be removed

erichuang27 (approved):
looks good

PR #95, opened 2023-02-12
Recommend challenges

cgu2020:
### Recommended challenges 

- [x] Longitude and latitude coordinates for events depending on the first challenge
- [x] gateway for retrieving recommended challenges based on geolocation


cli66 (approved):
Logic looks good, especially in the method getting recommended events based on distance.

server/src/event/event.service.ts
```diff
@@ -381,6 +392,12 @@ export class EventService {
   }
 
   async upsertEventFromDto(event: EventDto) {
+    const firstChal = await this.prisma.challenge.findFirst({
+      where: { id: event.challengeIds[0] },
+      select: { latitude: true, longitude: true },
+    });
+    console.log(firstChal);
```
youssef-attia, line 399: Do we want to have console.logs in the production code?

server/src/event/event.service.ts
```diff
@@ -224,6 +224,15 @@ export class EventService {
     });
   }
 
+  async getRecommendedEventsForUser(user: User, lat: number, long: number) {
```
youssef-attia, line 227: Maybe add the number of recommendations as a parameter for this function instead of it just being 4.

server/src/event/event.service.ts
neketka: No we do not

server/src/event/event.service.ts
neketka: I agree

admin/src/dto/event.dto.ts
```diff
@@ -30,6 +30,8 @@ export interface EventDto {
   defaultChallengeId: string;
   minimumScore: number;
   indexable: boolean;
+  latitude: number;
```
neketka, line 33: Make it more clear what this lat/long is, maybe firstLatitude instead.

server/prisma/schema.prisma
```diff
@@ -91,6 +91,8 @@ model EventBase {
   defaultChallengeId String          @unique
   trackers           EventTracker[]
   activeGroups       Group[]
+  latitude           Float
```
neketka, line 94: Same issue here

server/src/event/event.service.ts
```diff
@@ -224,6 +224,15 @@ export class EventService {
...
+    const evs: EventBase[] = await this.prisma.$queryRaw`
+      select * from "EventBase" ev 
+      order by ((ev."latitude" - ${lat})^2 + (ev."longitude" - ${long})^2)
```
neketka, line 230: If you do change the name remember to change it here too

neketka (changes requested):
Look at comments

BrianLa0616:
Great work! However, there is an issue that happens when we change the location of an event. I don't think that the change in location is actually registered in the challenge that it is in, although it is reflected in the event location.

neketka (dismissed):
lgtm

neketka (approved):
LGTM

PR #98, opened 2023-02-13
Added methods to manage organizations

BrianLa0616:
### Added methods to join organization and add managers to organizations.

- [x] implemented addManager and joinOrganization

Remaining TODOs:

- [x] allow adding managers from admin page


admin/src/components/Organizations.tsx
```diff
@@ -86,12 +86,14 @@ function fromForm(
...
     id,
     name: (form[0] as FreeEntryForm).value,
+    add_manager_email: form[1] ? (form[1] as FreeEntryForm).value : "",
   };
 }
 
 function toForm(group: OrganizationDto) {
   return [
     { name: "Name", characterLimit: 256, value: group.name },
+    { name: "Add Manager by Email", characterLimit: 256, value: "" },
```
neketka, line 96: I was thinking something along the lines of another dialog box to do this.

server/src/organization/organization.dto.ts
```diff
@@ -4,7 +4,9 @@ export interface OrganizationDto {
   accessCode: string;
   members: string[];
   events: string[];
+  managers: string[];
   defaultEventId: string;
+  manager_email: string;
```
neketka, line 9: DTO looks different here, also this does not belong here. Instead, the frontend should be calling addManager to do this

server/src/organization/organization.gateway.ts
```diff
@@ -61,11 +61,13 @@ export class OrganizationGateway {
...
-      const org = await this.orgService.upsertOrganizationFromDto(
-        data.organization as OrganizationDto,
-      );
+      const org_data = data.organization as OrganizationDto;
+
+      const org = await this.orgService.upsertOrganizationFromDto(org_data);
 
       await this.orgService.addAllAdmins(org);
+      if (org_data.manager_email !== '')
+        await this.orgService.addManager(user, org_data.manager_email, org);
```
neketka, line 70: Don't do this logic here

neketka (changes requested):

cli66:
The logic looks great to me! Would it be better to search for the manager using their user id instead of the email since the id is generated by us? Also is this fully ready to be tested in the frontend?

neketka (approved):

PR #97, opened 2023-02-12
Updated node version and added scripts to simplify db work

neketka:
### Summary 

This pull request updates the node version to the newest LTS and creates scripts that automate local testing of the db schema.

- [x] Push node to 18 LTS
- [x] Create dbreset

### Test Plan 

Make sure everyone can run these scripts

PR #99, opened 2023-02-14
fixed scripts to simplify db work

youssef-attia:
based on #97


erichuang27 (approved):
Great work finding this bug!

PR #100, opened 2023-02-14
Made changes to Flutter packages and deployment versions.

nirbhaysinghnarang:
### Summary

In this PR, I updated versions and deployment targets for the Flutter app to run on iOS simulators.
Also fixed a bug that requests location when another request is already running, causing the app to crash.

Note: 
App currently configured to run on iOS 15.2
Current version of Flutter: 3.3.x


neketka (approved):
LGTM

PR #102, opened 2023-03-03
Session Logging continued

youssef-attia:
### Summary 

A new table has been created to store logs of user events such as joining or leaving a group, logging in, creating deleting or editing users, changes related to events, challenges, and rewards, and more. This will make it easier to debug breakages or misuses in the system once it is published.

### Test Plan 

Run the server and trigger several different actions and then preview the logging table to insure it includes all needed information.


nirbhaysinghnarang:
Good work! I think the test plan looks solid, maybe sometime in the future we can set up an automated test suite so that we don't have to rely on manual testing?

neketka (approved):
lgtm

PR #105, opened 2023-03-06
Bug fix

BrianLa0616:
After our code clean up during the winter break, there was in issue with displaying group data on the admin front page.

This PR fixes that bug.


neketka (approved):
lgtm

PR #106, opened 2023-03-07
Group invitation system

youssef-attia:
### Summary 
Added a way to invite users to groups. The extent of the changes are just the backend. The frontend will need to add a popup to react to the invitation and depending on the reaction, the user will either join or ignore the group.

### Test Plan 

Tested directly using postman


server/src/group/group.gateway.ts
```diff
@@ -86,4 +87,14 @@ export class GroupGateway {
       await this.groupService.emitUpdateGroupData(group, false);
     }
   }
+
+  @SubscribeMessage('sendGroupInvite')
+  async sendGroupInvite(
+    @CallingUser() user: User,
+    @MessageBody() data: SendGroupInviteDto,
+  ) {
+    if (!user.administrator) return;
```
erichuang27, line 96: send a this.clientservice.emitErrorData() in the future

server/src/group/group.service.ts
```diff
@@ -306,6 +306,25 @@ export class GroupService {
...
+    const dto: GroupInviteDto = {
+      groupId: await (await this.dtoForGroup(group)).friendlyId,
+      username: username,
+    };
+
+    if (user) {
+      this.clientService.sendUpdate('groupInvitation', user.id, !!admin, dto);
+    } else {
+      this.clientService.sendUpdate('groupInvitation', group.id, false, dto);
+      this.clientService.sendUpdate('groupInvitation', group.id, true, dto);
```
erichuang27, line 324: Why, explain

erichuang27 (approved):

BrianLa0616:
Great work! I was wondering if we should also emit an update for the user that is receiving an invitation. It might not be necessary, but I think that it'll be a good addition for anyone listening to the user.

server/src/group/group.gateway.ts
```diff
@@ -86,4 +87,13 @@ export class GroupGateway {
...
     }
   }
+
+  @SubscribeMessage('sendGroupInvite')
+  async sendGroupInvite(
+    @CallingUser() user: User,
+    @MessageBody() data: SendGroupInviteDto,
+  ) {
+    const group = await this.groupService.getGroupForUser(user);
+    await this.groupService.emitGroupInvite(group, data.targetUsername, user);
```
neketka, line 97: Change data.targetUsername to user.username

neketka (changes requested):
All good except the username issue.

neketka (approved):
LGTM

PR #111, opened 2023-03-17
User management

BrianLa0616:
### Summary 

This pull request allows the management of users on the front end as well as banning users on the back end.

- [x] fixed old version of table
- [x] ban users that prevent them from logging in and showing up on leaderboards

Remaining TODOs:

- [ ] implement banning in the front end

### Test Plan 

Use postman and interact with admin page

### Notes 

We assume that all changes in the front end will go through in the back end, which may not be true in some rare cases.


youssef-attia:
Great work with this PR. The code for banning a user all looks good. I'm wondering if we have any plans to change the login function to return some sort of FailedLogin enum when the login fails so that we can tell the user why the login failed. There are cases where the user doesn't exist or the user is banned and we may want to differentiate between the response on the frontend depending on the reason.

neketka (approved):
LGTM

neketka (approved):
LGTM

PR #113, opened 2023-03-24
Added segmented registration flow

youssef-attia:
### Summary 

Registration needed to be split into two steps with one being basic information and the next being auxiliary information. This PR splits registration into those two steps.

### Test Plan 

Standard Postman testing.


server/src/auth/auth.service.ts
```diff
@@ -128,16 +129,17 @@ export class AuthService {
...
       user = await this.userService.register(
         idToken.email,
-        req.username,
-        req.year,
-        req.major,
         req.lat,
         req.long,
         authType,
         idToken.id,
+        req.username && req.year && req.major ? true : false,
```
erichuang27, line 139: does this line up with the isRegistered field? Should have a comment to make it clear

erichuang27 (commented):
Looks good other than that comment, up to you if you want to add

cgu2020 (commented):
Check user.service. dtoForUserData requires your isRegistered field. Not sure if others functions require it but make sure it builds.

neketka:
Nirbhay has discovered a simpler way of doing this without extra database work

PR #116, opened 2023-03-29
Challenge.service.spec.ts

cgu2020:
### Summary 

This pull request is for testing the challenge module.

- [ ] removeChallenge is not removing challenge due to the user not being the in the chal's event's org's managers (commenting the query runs the test fine)
- [ ] Is challenge completed by user doesnt work as expected. Test might be written wrong or the implementation is just bad
- [ ] Changing all absolute paths to relative paths as the tests are failing
- [ ] Adding to CICD


neketka:
Too old, this will be covered by another PR

PR #121, opened 2023-04-17
Compass and Footer Design

youssef-attia:
### Summary 

Designing compass and footer for the app from lo-fis

### Test Plan 

Running the app.


youssef-attia:
Closed for now because of redesigns

PR #123, opened 2023-04-28
Challenges Preview

BrianLa0616:
### Summary 

This PR allows users to preview challenges

- [x] implemented challenge previews
- [x] fixed location of close button

Remaining TODOs:

- [ ] connect start button


cathli66 (approved):
I tested out the feature and love how it looks, especially when you included the feedback about making the toggle blue! It would be more intuitive if you updated the points tag on the top of the preview to 18 when you toggle the +3 bonus points option. Also, the Explore button appears a bit off-center. Overall, this looks almost exactly like the Figma. Great work!

PR #124, opened 2023-04-28
Fix challenge dto, add placeholder challenge cell widget.

nirbhaysinghnarang:
### Summary 

In this PR, I fixed the Challenge DTO and Challenge Model on the Flutter app to be in line with the new DTO on the backend.
Also added a placeholder widget for the ChallengeCell so that there are no compile time issues.


neketka (approved):
LGTM

PR #125, opened 2023-05-02
User schema update

nirbhaysinghnarang:
### Summary 
- Replace major with username across backend services for user.
- Necessary changes on FE.


youssef-attia:
Great changes so far, the registration flow looks really nice. I'm wondering if we have some sort of plan for phasing out the old designs, I see that right now we're just commenting them out but we could likely delete them at some point since they're in the repo history.

PR #126, opened 2023-05-04
Connect Journeys page to backend

youssef-attia:
### Summary 
Connecting the journey page to the backend

### Test Plan 
Run the page and check it pulls correctly from backend


neketka:
Taken over by brian

PR #128, opened 2023-05-05
Gameplay page

erichuang27:
### Summary 

Created the gameplay page and started to create the map

Remaining TODOs:

Connect to backend

### Test Plan 

Test out the UI on android and Ios


BrianLa0616:
Great work on the gameplay page, especially the map!

cgu2020 (commented):
Map looks good! We should try caching a tile layer for the Ithaca area since most if not all challenges will be here.

neketka:
Taken over by chris

PR #131, opened 2023-09-17
Add testing infrastructure

neketka:
### Summary

This pull request is the first step towards implementing containerized unit and end-to-end testing.

- [x] Add unit test and e2e test runner to dockerfile
- [x] Add env variables to select unit and e2e tests 
- [x] Add script to run tests locally in a container
- [x] Add a simple integration test for the user service
- [x] Add testing infrastructure to CI/CD pipeline
- [x] Optimize dbreset script for better logging and performance 
- [x] Create a setup script that installs the package, sets up the database, and verifies the setup using e2e tests

Remaining TODOs:

- [ ] Implement more tests

### Test Plan 

Make sure infra runs on all computers

### DIFF
--- a/server/src/challenge/challenge.gateway.ts
+++ b/server/src/challenge/challenge.gateway.ts
@@ -5,10 +5,10 @@ import {
   WebSocketGateway,
 } from '@nestjs/websockets';
 import { User } from '@prisma/client';
-import { UserGuard } from 'src/auth/jwt-auth.guard';
-import { EventGateway } from 'src/event/event.gateway';
-import { GroupGateway } from 'src/group/group.gateway';
-import { UserGateway } from 'src/user/user.gateway';
+import { UserGuard } from '../auth/jwt-auth.guard';
+import { EventGateway } from '../event/event.gateway';
+import { GroupGateway } from '../group/group.gateway';
+import { UserGateway } from '../user/user.gateway';
 import { CallingUser } from '../auth/calling-user.decorator';
 import { ClientService } from '../client/client.service';
 import { ChallengeService } from './challenge.service';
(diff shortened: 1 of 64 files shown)


server/src/user/user.e2e-spec.ts
```diff
@@ -0,0 +1,44 @@
...
+  beforeAll(async () => {
+    moduleRef = await Test.createTestingModule({
+      imports: [UserModule],
+    }).compile();
+
+    app = moduleRef.createNestApplication();
+    await app.init();
+  });
+
+  it(`Create and check user`, async () => {
```
BrianLa0616: Great infrastructure! This will make it easy for us to do test-driven development!

server/src/challenge/challenge.gateway.ts
```diff
@@ -5,10 +5,10 @@ import {
   WebSocketGateway,
 } from '@nestjs/websockets';
 import { User } from '@prisma/client';
-import { UserGuard } from 'src/auth/jwt-auth.guard';
-import { EventGateway } from 'src/event/event.gateway';
-import { GroupGateway } from 'src/group/group.gateway';
-import { UserGateway } from 'src/user/user.gateway';
+import { UserGuard } from '../auth/jwt-auth.guard';
```
BrianLa0616, line 8: Is there a specific reason we are changing to relative path from absolute?

cgu2020 (changes requested):
The infrastructure looks good and works! I am getting 'should be defined' fails on unit testing which seems very familiar from last semester. I think I can port over my changes in the jest branch and hopefully those dependencies get resolved. Otherwise, the infrastructure is good.

server/src/challenge/challenge.gateway.ts
neketka: Yes, this is so that Jest can resolve modules properly since it executes from a different path than the regular server.

cgu2020 (approved):
Lgtm

PR #137, opened 2023-10-08
Finish connecting journey's page to backend

BrianLa0616:
### Summary 

Finishing connecting journey's page to backend

Remaining TODOs:

- [ ] Test using frontend (events are currently not showing up)

### Breaking Changes  

Changed @relation(player) to (orgToUser)


game/lib/journeys/journeys_page.dart
```diff
@@ -132,18 +144,95 @@ class _JourneysPageState extends State<JourneysPage> {
...
+                    1000,
+                    [
+                      EventRewardType.PERPETUAL,
+                      EventRewardType.LIMITED_TIME_EVENT
+                    ],
+                    false,
+                    false,
+                    false);
+              }
+              print("search results: " + myEventModel.searchResults.toString());
```
tuckerstanley, line 164: Consider removing before code gets merged in

game/lib/journeys/journeys_page.dart
```diff
@@ -132,18 +144,95 @@ class _JourneysPageState extends State<JourneysPage> {
...
+                final reward =
+                    event.rewardIds.length == 0 ? null : event.rewardIds[0];
+                final tracker = trackerModel.trackerByEventId(event.id);
+                final format = DateFormat('yyyy-MM-dd');
+                final chal = event.challengeIds.length == 0
+                    ? null
+                    : challengeModel.getChallengeById(event.challengeIds[0]);
+                final complete = tracker?.prevChallengeIds.length ==
+                    event.challengeIds.length;
+                final timeTillExpire =
```
tuckerstanley, line 182: Given that these variables are constantly defined due to the for loop should they be of type final?

tuckerstanley (commented):
Code looks good overall so far, a couple of comments though.

cgu2020 (commented):
I was not able to run the react server. It kept crashing after google login. My working branch was fine. I also couldn't see any data in the frontend maybe the backend was crashing/not populated.

`cornell-go-postgres-1  | 2024-02-12 03:16:26.939 UTC [58] LOG:  could not receive data from client: Connection reset by peer
cornell-go-server-1 exited with code 0`

game/lib/model/event_model.dart
```diff
@@ -11,12 +11,24 @@ class EventModel extends ChangeNotifier {
 
   EventModel(ApiClient client) : _client = client {
     client.clientApi.updateEventDataStream.listen((event) {
-      if (!event.event is String) {
-        searchResults = event.event;
+      if (!(event.event is String)) {
+        if (!_events.containsKey(event.event.id)) {
+          if (searchResults != null) {
```
neketka, line 16: This is not easy to follow as well but this could be by design of our API. I suggest an improvement (easy one) for the API to clean up this logic:

Currently, this is what 
UpdateEventDataDto {
  event: EventDto | string;
  deleted: boolean;
} 
looks like.

UpdateEventDataDto {
  event: EventDto | string;
  isSearchResult: boolean; // This will be true when requestAllEventData is called
  deleted: boolean;
} .

Now instead of searchResults being nullable, simply make it an empty list and clear it every time the frontend search function (the one up above) is called. Keep the deletion/update logic the same. But, whenever isSearchResult is true, simply add that event to the searchResults list. This will clean up the logic a lot.

neketka (changes requested):
Left a few comments related to clean code and some inconsistencies, but after those, this looks good to go. Good work!

game/lib/model/event_model.dart
BrianLa0616: I cleaned up my code in a different way without changing the dto. Hopefully it's a bit clearer now

game/lib/model/event_model.dart
```diff
@@ -11,12 +11,24 @@ class EventModel extends ChangeNotifier {
...
   EventModel(ApiClient client) : _client = client {
     client.clientApi.updateEventDataStream.listen((event) {
-      if (!event.event is String) {
-        searchResults = event.event;
+      if (!event.deleted) {
+        //initialize searchResults if null
+        searchResults = searchResults ?? [];
+        //add or update event
+        if (!_events.containsKey(event.event.id)) {
+          searchResults!.add(event.event);
```
neketka, line 19: The issue with this condition is if we request event data outside the search functionality it will suddenly appear there (which is not the case at the moment but this could cause a bug in the future).

game/lib/api/game_api.dart
```diff
@@ -140,37 +140,36 @@ class ApiClient extends ChangeNotifier {
...
-        _createSocket(false);
-        return loginResponse;
-      }
-      authenticated = false;
-      _clientApi.disconnectedController.add(null);
-      notifyListeners();
-
-      print("Failed to connect to server!");
-      return null;
+    // if (pos != null) {
```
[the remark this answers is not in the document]
neketka: This is not yet resolved.

game/lib/api/game_api.dart
```diff
@@ -140,37 +140,36 @@ class ApiClient extends ChangeNotifier {
...
+      _createSocket(false);
+      return loginResponse;
     }
+    authenticated = false;
+    _clientApi.disconnectedController.add(null);
+    notifyListeners();
+
+    print("Failed to connect to server!");
+    return null;
+    // }
```
neketka, line 172: Here as well

neketka (changes requested):
Still a couple of things left, but almost there!

game/lib/model/event_model.dart
neketka: You can fix this without touching the API at all by adding filtering inside the model. For example, if you have a filter in the frontend (arts quad only + non expiring + whatever), you can set it inside the model whenever searchEvents is called (along with the requestAllEvents call). On every _events update, rerun the filter over all the events and put all those inside searchResults.

BrianLa0616: Shouldn't the filtering be done on the widget page? Because otherwise based on the filtering the user does, the filter function in event_model will also change

neketka (approved):
Anything further is beyond the scope of this PR. I'm calling it done!

PR #138, opened 2023-10-18
Bottom Navigation Bar

cgu2020:
### Summary 

This pull request is the first step towards implementing bottom navigation bar

- [x] svg icons


tuckerstanley (changes requested):
Regarding the navigation changes, they look good, I tested it on the app, and it works. 

However, In order to get the app running though, I needed to make a couple changes due to errors. In the text below in `back_btn.dart`, I needed to change line 15 to `PlayAnimationBuilder<double>(` and then change 18 to `builder: (child, value, _) { `. Not sure if this was erroring for you as well, but it was preventing me from running the app. I also had to change the dependencies in pubspec.yaml, but might just be an issue on my system.

BrianLa0616:
Great work! I really like how you have not only a highlight for the current tab, but also how you have an animation that increases the text size!

tuckerstanley (approved):
Looks good, thanks for fixing up the dependencies to match the new versions. That was the only fix that was requested, the rest looks good!

PR #144, opened 2023-11-14
First steps in implementing new Achievement System & Schema

tuckerstanley:
### Summary 

This pull request is the first step towards implementing a new Acheivement System. In this PR so far I have implemented the new locationType and achievementType enums. I created a datastructure to define an achievement, and one to track it. I then created the nest module and started implementing the service, gateway, and module files. This PR is not complete, but is the start to implementing the achievements.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:18.14.0
+FROM node:20-alpine
 
 EXPOSE 80
 EXPOSE 8000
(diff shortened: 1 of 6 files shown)


Dockerfile
cgu2020, line 1: Is there any reason to update our Node to the latest version because this doesn't seem necessary.

cgu2020 (commented):
Besides the change to the docker image, everything else looks good.

PR #153, opened 2024-02-19
Integrating Google Maps API

cathli66:
### Summary 

Integrating Google Maps API into CornellGo's app for our gameplay. 
Includes pushing the API key (consider removing in the future as it should be secret information)

- [x] Updated Android's build files to include Google Maps API class
- [x] Changed gameplay_map.dart from Flutter map (old) to Google Maps (new)

Remaining TODOs:

- [ ] Update iOS side to work with Google Maps API and test
- [ ] Move maps into gameplay page
- [ ] Display and zoom to user's current location
- [ ] Draw some random circle around a challenge for user testing

### Blockers 

- Waiting for DTI Business Leads to check that the Maps SDK is enabled on Google Cloud Console because they set up the account and the map is not showing up

### DIFF
--- a/game/android/app/src/main/AndroidManifest.xml
+++ b/game/android/app/src/main/AndroidManifest.xml
@@ -6,7 +6,9 @@
   
     <meta-data
         android:name="com.google.android.geo.API_KEY"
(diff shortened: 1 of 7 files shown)


neketka:
Leaks api key

PR #161, opened 2024-03-02
Fix heroku failure

neketka:
### Summary 

This pull request is the first step towards getting heroku to successfully deploy.

- [x] Changed node versions to 20
- [x] Moved startup logic to /server/start.sh


cathli66 (approved):
Everything works on my end after changing the shebang. The CRLF -> LF line ending change doesn't get saved/committed though. Might be something we'll have to change in VSCode each time? We can look into it later.

PR #164, opened 2024-03-05
Change CRLF to LF line endings

neketka:
### Summary 

Change CRLF to LF line endings


cathli66 (approved):
Line endings got updated for me!

PR #165, opened 2024-03-07
Added settings profile page

3TTemi:
### Summary 

This pull request adds the settings page which is accessed by other pages
Page includes references to editing profile, notifications, feedback, and logging out

### Notes 

Only static, none of the options on the page are dynamic/navigators


game/lib/profile/settings_page.dart
```diff
@@ -0,0 +1,134 @@
...
+                            'assets/icons/head.svg',
+                          ),
+                          title: Text(
+                            'Edit Profile',
+                            style: TextStyle(fontFamily: 'Poppins'),
+                          ))),
+                  Divider(height: 1),
+                  LayoutBuilder(builder:
+                      (BuildContext context, BoxConstraints constraints) {
+                    return Container(
```
cathli66, line 59: You can follow this format for each individual button, but put the entire column with all buttons inside the layoutbuilder if you keep it!

cathli66 (changes requested):
Good work so far! See comments 😄

cathli66 (approved):
Looks like the Figma designs, nice work! I also reverted some iOS build files that were causing merge conflicts.

PR #170, opened 2024-03-11
Cl893/photos

cathli66:
### Summary 

This pull request is the first step towards implementing the camera and photo taking feature after a user has completed a challenge.

- [x] Connect UI with camera
- [x] Button to take picture and display on the frontend page

Remaining TODOs:

- [ ] Finish UI from Figma
- [ ] Backend support for images
- [ ] Connect page with backend: sending image file path to backend to store

### Breaking Changes  

- Database schema change: adding support for user photos (still in the works, not pushed yet)


game/lib/gameplay/camera_page.dart
```diff
@@ -0,0 +1,168 @@
...
+            break;
+          default:
+            // Handle other errors here.
+            break;
+        }
+      }
+    });
+  }
+
+  toggleCamera() async {
```
neketka, line 53: Add a return type here Future<...> otherwise this returns dynamic

game/lib/gameplay/camera_page.dart
```diff
@@ -0,0 +1,168 @@
...
+              ],
+            ),
+            Text(
+              "Remember this moment with a photo!",
+              style: TextStyle(
+                fontSize: 18,
+                fontWeight: FontWeight.w700,
+              ),
+            ),
+            imagePath == ""
```
neketka, line 142: It looks like you can use a better state here instead of imagePath == "". It's confusing to see why an empty image path means we have a camera preview. If you plan to keep this, please add a comment explaining why you make this comparison.

neketka (changes requested):
Solid start, some minor changes requested

PR #171, opened 2024-03-11
Initial commit for connected challenges

BrianLa0616:
### Summary 

Connect challenges to backend

- [x] Connected location and points to backend
- [x] Added locations and points to preview

Remaining TODOs:

None


cgu2020 (commented):
Location type in the admin front page should be a drop down menu. Was having trouble creating events because of the location type. Difficulty in the backend is easy, normal, hard but is easy, medium, hard on the frontend. We should change the backend to reflect that (not part of PR just point it out). Journeys page not showing any cells at all. I am not sure if thats intentional or not.

cathli66 (approved):
Tested on my end as well and it works! I agree with Chris about the dropdown instead of text field and making the backend consistent (normal -> medium), but both are not too important for now.

PR #174, opened 2024-03-23
Byl8/challenges backend documentation

BrianLa0616:
### Summary 

Added comments for challenges model on flutter and removed dummy data.


cathli66:
Thanks for adding more documentation for the backend methods! I was actually finished with connecting the Preview with the correct fields in my own PR... we should really communicate to each other when we're doing extra tasks, that's my bad!

BrianLa0616:
> Thanks for adding more documentation for the backend methods! I was actually finished with connecting the Preview with the correct fields in my own PR... we should really communicate to each other when we're doing extra tasks, that's my bad!

cathli66:
Another quick comment - we should use the LocationType enum more and store location as that type. Try to only convert between LocationType and string on the client facing ends (admin and game) rather than storing them as strings everywhere in the backend. Since our current codebase already has location as a string, we can treat this a task for later.

BrianLa0616:
> Another quick comment - we should use the LocationType enum more and store location as that type. Try to only convert between LocationType and string on the client facing ends (admin and game) rather than storing them as strings everywhere in the backend. Since our current codebase already has location as a string, we can treat this a task for later.

I'm waiting for Nikita's PR on the filtering because he already has that implemented and we'll have to fix it anyways after he pushes it

cathli66:
> > Another quick comment - we should use the LocationType enum more and store location as that type. Try to only convert between LocationType and string on the client facing ends (admin and game) rather than storing them as strings everywhere in the backend. Since our current codebase already has location as a string, we can treat this a task for later.
> 
> I'm waiting for Nikita's PR on the filtering because he already has that implemented and we'll have to fix it anyways after he pushes it

Just spoke to Nikita and he might redo his PR since you already added in a lot of the fields he needs. I still need to add in category into EventBase, so I can edit location to be LocationType in the same PR. Either way your current PR is fine :)

game/lib/challenges/challenges_page.dart
```diff
@@ -136,18 +106,96 @@ class _ChallengesPageState extends State<ChallengesPage> {
                     ],
                   ),
                 ),
-                Expanded(
-                  child: ListView.separated(
-                    padding: const EdgeInsets.all(0),
-                    itemCount: cells.length,
+                Expanded(child: Consumer4<EventModel, GroupModel, TrackerModel,
```
neketka, line 109: It looks like this code is a repeat of what the master branch has. How is this different than the code we have currently?

game/lib/journeys/journeys_page.dart
```diff
@@ -159,15 +175,15 @@ class _JourneysPageState extends State<JourneysPage> {
...
                                 Image.network(
                                     "https://picsum.photos/250?image=9"), // dummy data for now; should pass in thumbnail parameter
                                 event.description,
                                 locationCount,
                                 numberCompleted,
                                 complete,
                                 difficulty,
-                                event.minimumScore,
-                                0),
+                                total_points),
```
neketka, line 186: Nit: camelCase it

server/src/challenge/challenge.e2e-spec.ts
```diff
@@ -167,7 +169,9 @@ describe('ChallengeModule E2E', () => {
       const secondChalDTO: ChallengeDto = {
```
neketka, line 169: Nit: Dto instead of DTO

server/src/challenge/challenge.e2e-spec.ts
```diff
@@ -127,7 +127,9 @@ describe('ChallengeModule E2E', () => {
       const chaldto: ChallengeDto = {
```
neketka, line 127: Nit: chalDto

neketka (changes requested):
Overall good, some minor nits

game/lib/challenges/challenges_page.dart
BrianLa0616: It isn't. I just forgot to add documentation and I didn't create a new branch.... oops

cathli66 (approved):
Ok everything looks good! Thanks for fixing the difficulty tag on the frontend too

neketka (approved):
Addressed all my comments!

PR #183, opened 2024-04-12
Byl8/profile backend

BrianLa0616:
### Summary 

Connected the profile page to the backend

- [x] Added user score from backend
- [x] Removed effect of name in the registration flow (hasn't been completely removed)
- [x] Retrieve completed events from the backend and display 2 of them on the profile page

Remaining TODOs:

- [ ] Finish connecting the rest of the profile page to the backend
- [ ] Update frontend UI as it has changed again

### Test Plan 

Manually added completedChallenges to the database using prisma studio. Unsure how to test otherwise as complete challenges are user specific


game/lib/model/tracker_model.dart
```diff
@@ -8,6 +8,7 @@ class TrackerModel extends ChangeNotifier {
 
   TrackerModel(ApiClient client) : _client = client {
     client.clientApi.updateEventTrackerDataStream.listen((event) {
+      print("TrackerModel: updateEventTrackerDataStream");
```
neketka, line 11: Delete this debug info

neketka (changes requested):
Code is mostly good, but I had some issues with the profile page. When I navigate to the completed challenges, then back, the nav bar disappears and when I go to swipe back on more time it brings me back to the completed challenges page. The likely reason is that instead of doing a Navigator.pop() you add the profile to the stack (which would cause this exact behavior).

game/lib/details_page/details_page.dart
```diff
@@ -51,11 +51,9 @@ class _DetailsPageWidgetState extends State<DetailsPageWidget> {
                     TextFormField(
                       decoration: const InputDecoration(
                         border: OutlineInputBorder(),
-                        labelText: 'e.g. Jane Doe',
+                        labelText: 'e.g. Jane Doe (will remove I think?)',
```
neketka, line 54: This is not production-ready

neketka (changes requested):

game/lib/details_page/details_page.dart
neketka: I think cathryn fixes it tho so ignore

server/src/challenge/challenge.gateway.ts
```diff
@@ -93,7 +93,7 @@ export class ChallengeGateway {
       );
 
       await this.groupService.emitUpdateGroupData(group, false);
-      await this.eventService.emitUpdateEventTracker(tracker);
+      await this.eventService.emitUpdateEventTracker(tracker, user);
```
neketka, line 96: I think let's remove ", user" cuz it should propagate the update to all ppl

server/src/challenge/challenge.gateway.ts
```diff
@@ -72,7 +72,7 @@ export class ChallengeGateway {
...
```
neketka, line 75: I think let's remove ", user" cuz it should propagate the update to all ppl

game/lib/details_page/details_page.dart
neketka, line 54: On second thought don't commit this to the repo even if it will be fixed later.

neketka (changes requested):
Address small comments and good to go!

neketka (approved):
Addressed all my concerns!

PR #184, opened 2024-04-14
Nmp79/implementing filtering

nit0107:
### Summary 

This pull request is related to the filtering functionality of challenges/journeys. When a user selects certain filters on the filterForm, the search will result only displays events that match those filters. 

- [x] Added state values for FilterForm to track selected options and passed them to/from the Challenge form
- [x] Defaults for location/category is empty selections, difficulty level is "easy" (as per the design handoff Figma file)
- [x] Implemented logic in ChallengeForm to not display events where the selected Difficulty and Locations do not match
- [x] Implemented Clear function to remove previous selections & view all challenges 

Remaining TODOs:

- [ ] fully implement in backend 

### Test Plan 

https://github.com/cornell-dti/cornell-go/assets/68392119/c1fb7321-75e8-4aea-af4a-97b9c43b6bfc


cgu2020 (commented):
Just suggestions. The filter defaults too easy difficulty (not sure if we should clear all filters when the app opens). The difficulty section also is not togglable like the other categories. Location names look like how they are in the DB which we should parse into actually names. Otherwise, good work, ui looks good!

neketka (approved):
Lgtm, but resolve conflicts

cathli66 (approved):
Everything overall looks great and works as expected! I went ahead and fixed some minor errors with the height of the home nav bar and search bar. You hard coded it the height in pixels which didn't look right on my screen! I also changed filtering such that it takes into account search and filters together (you had the filters only apply when the search bar is empty).

PR #199, opened 2024-04-21
Byl8/challenge bug fix

BrianLa0616:
### Summary 

Fix simple bug with names

- [x] Challenge challenge name to be event name

### Test Plan 

Screenshot 2024-04-21 at 3 21 08 PM


cathli66 (approved):
friendlyLocation should probably be like friendlyDifficulty and include the enum values - I'll fix in my PR!

PR #200, opened 2024-04-21
Fix overly permissive regex range security issue

neketka:
Addresses [github.com/cornell-dti/cornell-go/security/code-scanning/1](http://github.com/cornell-dti/cornell-go/security/code-scanning/1)


cathli66 (approved):
LGTM

PR #205, opened 2024-04-24
Relax first challenge restriction

neketka:
Allow players to now receive contents of first challenge due to frontend bug. Change corresponding e2e test to match new behavior.


cathli66 (approved):
LGTM

PR #206, opened 2024-04-24
Fix socket.io room messaging leak

neketka:
### Summary 

This pull request attempts to solve the message leaking from sending messages to client. This issue not only causes security issues but also interferes with the normal usage of the app.

- [x] Fixed bug

Remaining TODOs:

### Test Plan 

Run with multiple users of different privileges and see if privileges seem to leak from one user to another.


cathli66 (approved):
Tested on prod, looks good!

PR #208, opened 2024-04-25
Ta375/leaderboard trophies

3TTemi:
### Summary 

Adds podium trophies to leaderboard display page 

- [x] Remade the podium_widgets.dart file, Instead of 6 widgets this is not condensed to one that dynamically changes the image based on the user position  (This also allows for general easier positioning and changes for the future) 

### Test Plan 

(Photo with some test data in the actual leaderboard scroll view - Can add sample data for testing)
image


game/lib/global_leaderboard/podium_widgets.dart
```diff
@@ -15,89 +23,63 @@ var pointsStyle = TextStyle(
...
-  return Center(
-      child: Container(
-          width: 106,
-          height: 62,
-          child:
-              Stack(alignment: FractionalOffset(0.5, 0.75), children: <Widget>[
-            SvgPicture.asset(
-              svgAssetPath,
-              semanticsLabel: '3rd Podium',
-            ),
```
cathli66, line 90: Your previous way of switching between highlighted and not podiums is a lot more efficient than this new mapping you have where you import svgs of all 3 podiums + trophies + labels. Is there a reason you couldn't add each individual trophy on top of your existing cylinders?

game/lib/global_leaderboard/podium_widgets.dart
```diff
@@ -15,89 +23,63 @@ var pointsStyle = TextStyle(
...
+          child: FittedBox(
+            alignment: Alignment.center,
+            fit: BoxFit.scaleDown,
+            child: Text(scoreList[0].toString() + " PTS", style: pointsStyle),
+          ),
+        ),
+      ),
+      Positioned(
+        right: 27,
+        bottom: 6,
```
cathli66, line 72: I'm a little concerned about these hard coded positioning values. They looked fine to me on an Android emulator but I'm quite sure it would look off for different phones.

game/lib/global_leaderboard/podium_widgets.dart
```diff
@@ -15,89 +23,63 @@ var pointsStyle = TextStyle(
...
+        child: SizedBox(
+          width: 60,
+          height: 29,
+          child: FittedBox(
+            alignment: Alignment.center,
+            fit: BoxFit.scaleDown,
+            child: Text(scoreList[2].toString() + " PTS", style: pointsStyle),
```
cathli66, line 79: Rather than trying to hard code the position of the points text inside the rounded rectangle label, you should create a rounded rectangle inside flutter and put the points inside that container. This guarantees that the points will never be off center of the label.

game/lib/global_leaderboard/global_leaderboard_widget.dart
```diff
@@ -83,132 +93,129 @@ class _GlobalLeaderboardWidgetState extends State<GlobalLeaderboardWidget> {
...
-                                    child: leaderBoardCell(
-                                      context,
-                                      user.username,
-                                      position++,
-                                      user.score,
-                                      user.userId == myUserModel.userData?.id,
+                      ),
+                      child: Container(
+                        width: 283.05,
+                        height: 432.0,
```
cathli66, line 177: Set the width and height to be a percentage of the total screen instead of hard coding values. For example, if you want the width to take up 90% of the screen width, you can do width: MediaQuery.of(context).size.width * 0.9.

game/lib/global_leaderboard/global_leaderboard_widget.dart
```diff
@@ -83,132 +93,129 @@ class _GlobalLeaderboardWidgetState extends State<GlobalLeaderboardWidget> {
...
+                        child: ListView(
+                          shrinkWrap: true,
+                          scrollDirection: Axis.vertical,
+                          children: [
+                            for (LeaderDto user in list.skip(3))
+                              Padding(
+                                padding: const EdgeInsets.only(
+                                    left: 30.95, right: 30.95, top: 16.0),
```
cathli66, line 185: I'm guessing you are adding this padding to center the container? You should wrap your Container child in a Align widget to center it instead. This is why the leaderboard is not centered when we tested it on our deployed iOS devices!

cathli66 (changes requested):
Take a look at the comments please! You should essentially only import the following svgs separately to work with:
3 trophies
3 podiums (red)
3 podiums (orange) - actually, you don't even need these. You should be able to set the color of the red podiums to orange in the code! 

So either 6 or 9 total svgs. Please remove the rest! It makes our app size unnecessarily larger.

cathli66 (approved):
Addressed all my comments!

PR #209, opened 2024-04-25
Allow skipping registration when google account already exists

neketka:
### Summary

This pull request allows skipping the registration process when a user has registered before.

- [x] Reorganize login code
- [x] Run dart fix --apply
- [x] Log into the main screen if initial login request succeeds
- [x] Add format all
- [x] Fix room leaking issue (again)
- [x] Realtime leaderboard


cathli66 (changes requested):
Thanks for adding the formatall shortcut and changing the app icon! The disconnect logic all works as expected. For leaderboard updates, points update in real time but the username changes don't reflect until I restart the app. Can't check if Google login works but I trust you

cathli66 (approved):
Addressed my comments, thank you!

PR #212, opened 2024-04-28
jxl9/achievements tracking logic

jasmineee-li:
### Summary 

Implementing achievements tracking backend
- [x] added achievement tracker CRUD functions
- [x] added logic to completeChallenge with checkAchievementProgress

Remaining:

- [ ] finish writing e2e tests
- [ ] debug


server/src/achievement/achievement.service.ts
```diff
@@ -194,4 +198,153 @@ export class AchievementService {
...
+    const ability = await this.abilityFactory.createForUser(user);
+
+    // find all achievements associated with the challenge that are accessible
+    // by user and have incomplete trackers; joins tracker to resulting query
+    const achs = await this.prisma.achievement.findMany({
+      where: {
+        OR: [
+          { linkedEventId: challengeId }, // achievements linked to the specific event of the challenge
+          { linkedEventId: null }, // achievements not linked to any specific event
+        ],
```
neketka, line 285: I like where this is going, make sure to include the various enums like achievement type and locationtype = ANY

server/src/achievement/achievement.service.ts
```diff
@@ -194,4 +198,153 @@ export class AchievementService {
...
+        AND: [
+          accessibleBy(ability, Action.Read).Achievement,
+          {locationType: curChallenge.location},
+        ],
+      },
+      include: {
+        trackers: {
+          where: {
```
neketka, line 293: This will only include trackers where this is true, but you don't want to read the achievements with trackers where this is true at all either, so include it in the above where condition instead.

server/src/achievement/achievement.service.ts
```diff
@@ -194,4 +198,153 @@ export class AchievementService {
...
+
+          
+      if (ach.achievementType === AchievementTypeDto.TOTAL_POINTS) {
+        tracker.progress += curChallenge.points;
+        if (tracker.progress >= ach.requiredPoints) {
+          // ach is newly completed; update tracker with completion date
+          tracker.dateComplete = new Date();
+        }
+      } else if (journeyOrChalAchShouldProgress) {
+        tracker.progress += 1;
```
neketka, line 335: Please use the prisma increment so we don't have a race condition

server/src/challenge/challenge.service.ts
```diff
@@ -150,6 +158,20 @@ export class ChallengeService {
       user.id,
     );
 
+    // check if the challenge is part of a journey
+    const isJourney = (await this.prisma.prevChallenge.count({
```
neketka, line 162: This should say is journey completed

server/src/challenge/challenge.service.ts
```diff
@@ -150,6 +158,20 @@ export class ChallengeService {
...
+      where: {
+        userId: user.id,
+        challengeId: eventTracker.curChallengeId,
+        trackerId: eventTracker.id,
+      },
+    })) === (await this.prisma.eventTracker.count({
```
neketka, line 168: This query always returns 1 so you likely meant something else

neketka (changes requested):
Good start!

neketka:
Moved to another pr

PR #213, opened 2024-04-28
Updates to gameplay map + more

cathli66:
### Summary 

This pull request implements some fixes to gameplay map and page. It also adds in the 3 new location types that PMs wanted when we discussed with dev leads about the dev social challenges.

- [x] User location icon now rotates from center and not bottom
- [x] Edited description and name text fields for gameplay page and completed challenge page
- [x] Completing the challenge now occurs upon pressing I've Arrived (before was Point Breakdown)
- [x] Removed Leave button after pressing I've Arrived to complete a challenge
- [x] Fixed hint radius initialization
- [x] Added 3 new location types (VET_SCHOOL, CENTRAL_CAMPUS, CORNELL_ATHLETICS)

Remaining TODOs:

- [x] User location icon resize and higher res
- [ ] After “I’ve Arrived”, Nevermind and Use Hint button are too close
- [x] After you use all hints, hint button should not appear on “Nearly There” popup

### Test Plan 

1. Start a challenge. Use two or three hints and try to keep track of how big the circle is, then leave challenge. Go back to the challenge, and check that the hint circle is the same size as how u left it. (The center of the circle may be different though)
2. Check that the user location icon is around the right size (should be a bit smaller than the hint/recenter buttons)
3. Upon completing, check that the challenge name appears in the Congratulations! dialog and that this is different from the challenge description on the top of the gameplay page.

Already tested on other devices during work session:
- User location icon rotates around center

### Breaking Changes  

- Database schema change (adding in 3 new location types)


neketka (approved):
Discussed over slack lgtm

PR #220, opened 2024-05-02
Fix challenge + join org issue

neketka:
### Summary 

Users can now see all challenges but not necessarily lat or long. Joining an org should allow you to see achievements of that org too.


cathli66 (approved):
New rule to relax read challenge permissions looks correct to me

PR #221, opened 2024-05-02
Cl893/fix race conditions

cathli66:
### Summary 

This pull request fixes some last minute bugs encountered after deployment! 

- [x] Fix race condition with I've Arrived modal popping up before completed challenge is called in the backend
- [x] Fixed bug where achievement trackers were not properly added for users
- [x] Minor UI changes to look good on iOS


BrianLa0616:
Is there a specific reason why you changed the return of completeChallenge from boolean to string? Does getting the name from a challenge not work?

BrianLa0616 (approved):
GOAT Cathryn. Hopefully this doesn't break stuff 😭 

We probably should have enums for difficulties instead of strings. Will work on that.

PR #222, opened 2024-05-05
some comments

BrianLa0616:
### Summary 

- [ ] some documentation


3TTemi:
Looks good, would be useful to include some tags

3TTemi:
Looks good, however it would be useful to include some tags like @param or @example so that devs understand how to call/use the components.

PR #224, opened 2024-09-22
Ranking List Names Updated with Ellipsis

3TTemi:
### Summary 

Fixes one of the low priority bugs  of overflowing with long useranems

Remaining TODOs:

- [ ] Create system to reduce points text in leaderboard (Ex: 15,000 -> 15k)

### Test Plan 

Can be seen in on the global leaderboard page with any longer test names or current names in the backend


BrianLa0616 (changes requested):
Use built in Text handler

game/lib/widget/leaderboard_cell.dart
```diff
@@ -78,7 +85,8 @@ Widget leaderBoardCell(
                       padding: const EdgeInsets.only(left: 10, right: 5),
                       child: Align(
                         alignment: Alignment.centerLeft,
-                        child: Text(name, style: nameStyle),
+                        child: Text(truncateWithEllipsis(name, 10),
```
BrianLa0616, line 88: This probably works fine, but I do think that it would be better to use Flutter's built in Text overflow handler with "overflow: TextOverflow.ellipsis". You might need to add some wrapper widgets though.

game/lib/widget/leaderboard_cell.dart
```diff
@@ -34,6 +34,13 @@ Widget leaderBoardCell(
     height: 0,
   );
 
+// Helper Function with ellipses functionality such that we can set max character length length/cutoff
+  String truncateWithEllipsis(String text, int maxLength) {
```
Shengle-Dai, line 38: I think it'll be better to truncate more characters if the original text just exceeds maxLength. This will make users less likely to be confused by unnecessary truncation of short texts.
For example: if the text is "I like that", this original implementation will truncate it into "I like tha...", but when users click to expand this truncation, they will only see one character "t" being truncated, and they might be sad of seeing only one letter.

Shengle-Dai (commented):
It's good overall, but as Brian mentioned, maybe there are more concise built-in methods to achieve the truncation!

game/lib/widget/leaderboard_cell.dart
```diff
@@ -34,6 +34,13 @@ Widget leaderBoardCell(
...
+    return (text.length <= maxLength)
+        ? text
+        : '${text.substring(0, maxLength)}...';
```
shah-esha, line 41: This looks great! However, it may be worthwhile to consider the event where the text has emojis or other special characters. Since these can take up more than one character, truncating them could cause the character to break and produce an odd string of text.

BrianLa0616 (approved):
Addressed comments! Good job!

PR #225, opened 2024-09-29
Byl8/debug xcode

BrianLa0616:
### Summary 

Fix dependency issues with Xcode and admin front end
- [x] Updated socket version
- [x] Updated ios version and google maps dependencies


3TTemi (approved):
Dependency updates look good and consistent. Was also able to run locally on my computer. Does Pubspec.lock being commented mean that we should all have the same dependencies now if we pull?

PR #228, opened 2024-10-08
Ta375/arrived race conditions

3TTemi:
### Summary 

This pull request handles race conditions that was most caused by spamming of the I've arrived button for challenges. 

Problem was in Internal testing causing users to get more than the maximum points for the challenge

https://www.notion.so/cornell-go/1dab9f87f80b46cf92d4b61e769b7639?v=f246714d0a2b43eba2bfb0466fa7af77&p=0009d2e0093a48fdab6260011190b91e&pm=s

Remaining TODOs:

- [ ] Some further testing when TestFlight is back up

### DIFF
--- a/game/lib/gameplay/gameplay_map.dart
+++ b/game/lib/gameplay/gameplay_map.dart
@@ -23,6 +23,25 @@ import 'package:game/model/group_model.dart';
 import 'package:game/model/event_model.dart';
 import 'package:game/model/challenge_model.dart';
 
+/// GameplayMap Widget
+///
+/// Displays an interactive map for gameplay, showing the user's location,
+/// hint circles, and challenge-related information.
+///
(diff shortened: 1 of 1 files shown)


Shengle-Dai (commented):
Introducing a boolean variable is a nice solution for the race condition! 
Also, I think it'll be more robust if we restrict backend logic to verify completed challenges as well.

shah-esha:
This is a really simple and effective solution to the race condition! It may also be beneficial to add a check within the onPressed functionality to make sure the user's score does not surpass the maximum.

BrianLa0616 (approved):
Great work!

I tested it by spamming the button while having print statements, and it only checks/sends for completion once. Hopefully this is robust enough and we don't need to also have it in the backend.

PR #229, opened 2024-10-27
Implement Rotating Cursor Feature with Dependency Fixes

Shengle-Dai:
### Summary 

Implemented the rotating cursor feature—updating the direction of the marker/cursor based on the user's magnetic heading direction.

- [x] fixed dependency issues on the local machine
- [x] implemented the core functionality to update the cursor rotation with the device's magnetic heading.

### Test Plan 
- [x] Simulate different compassHeading values to verify the cursor’s update.
- [ ] Test the feature on TestFlight to ensure performance.

### DIFF
--- a/game/ios/Podfile.lock
+++ b/game/ios/Podfile.lock
@@ -48,6 +48,8 @@ PODS:
     - GoogleUtilities/UserDefaults (~> 7.8)
     - PromisesObjC (~> 2.1)
   - Flutter (1.0.0)
+  - flutter_compass (0.0.1):
+    - Flutter
   - flutter_config (0.0.1):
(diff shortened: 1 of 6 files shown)


3TTemi:
Looks good! I think this solution will work. 
However, I would suggest adding a check for null values or defaulting the rotation to currentLocation.heading, as was used before. Additionally, I noticed that you didn't reference FlutterCompass in the dispose function. I'm not entirely sure how flutter_compass works, but if it streams data, it might need to be canceled or disposed of in that function as well.
Lastly, make sure that in the final PR, the mock data is either removed or moved to a separate function or utility file.

PR #230, opened 2024-10-28
Fix Journeys Search Bar Bug

shah-esha:
### Fix Journeys Search Bar Bug 

This pull request is to fix the Journeys search bar bug, where no results would show for any searches other than "e".

Remaining TODOs:

- [x] test changes on default journeys


BrianLa0616 (approved):
Good work!

We might want to look into having an EventsPage that JourneysPage and ChallengesPage both extends as there is a lot of commonalities between these two functions.

PR #231, opened 2024-11-06
Make progress bar width dynamic based on digit count and shorten larg…

Jcheng777:
…e numbers

- Adjusted progress bar width to dynamically scale based on the number of digits in total points, achievements, etc., providing better visual alignment.
- Implemented formatting to shorten large numbers (e.g., 1500 becomes 1.5K) for improved readability.

### Summary 

- [x] implemented X
- [x] fixed Y

Remaining TODOs:

- [ ] resolve bug 1
- [ ] implement Z

Depends on #{number of PR}

### Blockers 

### Breaking Changes  

### DIFF
--- a/game/lib/widget/progress_bar.dart
+++ b/game/lib/widget/progress_bar.dart
@@ -12,12 +12,33 @@ class LoadingBar extends StatelessWidget {
     this.totalTasks,
   );
 
+  String formatNumber(int number) {
+    if (number < 1000) {
+      return number.toString();
+    } else {
+      return (number / 1000).toStringAsFixed(1) + 'K';
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


Shengle-Dai (commented):
Nice try to shorten the number if it is greater than 1000 and make the width to have dynamic width!

However, I think it'll be better to make the progress bar length consistent for all achievements (maybe keep the 0.43 as the multiplier). Also, remember to follow the PR template and remove the unused section!

BrianLa0616 (changes requested):
When I have achievements as 100,000 points, there's a number format exception that occurs.

image

I also think that the points are center here. It should be closer to the right.
Screenshot 2024-11-18 at 3 02 29 PM

PR #232, opened 2024-11-11
Update journeys and challenges search bar

shah-esha:
### Summary 

This pull request implements additional and fixes current search features in both Journeys and Challenges.

- [x] implemented searching by difficulty
- [x] simplified code

Remaining TODOs:

- [x] update documentation for both Journeys and Challenges
- [x] test through simulation
- [ ] testing outside of just simulating the app locally


game/lib/journeys/journeys_page.dart
```diff
@@ -160,54 +170,66 @@ class _JourneysPageState extends State<JourneysPage> {
 
                       final challengeLocation = challenge.location?.name ?? "";
                       final challengeName = challenge.name ?? "";
+                      final eventDifficulty = event.difficulty?.name ?? "";
+                      final String eventName = event.name ?? "";
```
3TTemi, line 174: Could these be named Journey Difficulty and Journey Name for easier clarification of what they are referring to (since an event can almost be a journey or challenge depending on the number of challenges)

3TTemi:
Nice job on this fix! Make sure to remove the print debug statements, this should be done before you ever make a PR. Also for better readability, try and revisit some of the variable names, moving away from "event.examplefield" to ensure it's clear which entity the difficulty, name, and other fields refer to (see my comment in the journey code for reference).

BrianLa0616 (approved):
In a specific PR regarding addressing both cleaning up challenges and journeys, I would have a separate util function that can handle both the matching/searching feature in the journeys and challenges page. This can be done in another PR though.

Jcheng777 (commented):
Hey Esha! Great commit! I think the if-else statements could be cleaned up a little to avoid repetition. Otherwise really great work!

PR #234, opened 2024-11-16
Fix New User Google Login/Register Issue

Shengle-Dai:
### Summary 

This pull request fixes the problem when users cannot log in/register on the simulator if their information is not in the database.

- [x] Revised the logic of printing "unable to register user"
- [x] Added a logic of checking the user's existence in the database before directly logging in as a user

Dependency fixes are the same as <a>https://github.com/cornell-dti/cornell-go/pull/229</a>

### Test Plan 
- [x] Tested on ios simulator

### Notes 

If we call Geolocator.getCurrentPosition() twice on the simulator, it will stuck forever. (in geopoint.dart)

### DIFF
--- a/game/ios/Runner/AppDelegate.swift
+++ b/game/ios/Runner/AppDelegate.swift
@@ -4,7 +4,7 @@ import Firebase
 import GoogleMaps
 import flutter_config
 
-@UIApplicationMain
+@main
 @objc class AppDelegate: FlutterAppDelegate {
   override func application(
     _ application: UIApplication,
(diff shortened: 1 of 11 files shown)


BrianLa0616 (changes requested):
Your changes allow for relogin. However, there's now an issue where an existing user is redirected to the registration change every time. 

Specifically, this is because of the error that is thrown "flutter: Error occurred while checking user: FormatException: Unexpected character (at character 1)"

BrianLa0616 (approved):
Addressed comments. Good work!

PR #236, opened 2024-12-08
Jc2755/refactor challenges journeys

Jcheng777:
### Summary 

- [x] implemented X
- [x] fixed Y

Remaining TODOs:

- [ ] resolve bug 1
- [ ] implement Z

Depends on #{number of PR}

### Blockers 

### Breaking Changes  

### DIFF
--- a/game/lib/widget/progress_bar.dart
+++ b/game/lib/widget/progress_bar.dart
@@ -12,6 +12,17 @@ class LoadingBar extends StatelessWidget {
     this.totalTasks,
   );
 
+/**
+ * Formats a number into a readable string with a "K" suffix for thousands.
+ * 
+ * Converts numbers to a string format using different precision based on their size to ensure readability in UI displays.
+ * For numbers between 1,000 and 9,999, one decimal is shown; numbers exactly at 1,000 or 10,000 are shown without decimals.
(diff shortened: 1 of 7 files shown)


game/lib/widget/progress_bar.dart
```diff
@@ -12,64 +12,111 @@ class LoadingBar extends StatelessWidget {
...
-                      constraints.maxWidth,
-                  height: 13,
-                  alignment: Alignment.centerLeft,
-                  child: Container(
+                  Container(
+                    width: (totalTasks > 0 ? tasksFinished / totalTasks : 0) *
+                        constraints.maxWidth,
+                    height: 13,
+                    alignment: Alignment.centerLeft,
+                    child: Container(
```
3TTemi, line 75: Instead of having a container with a another container inside of it, the BoxDeocration should be able to be applied to the outer container.

3TTemi:
Nice job abstracting out repetitive code. I still see a lot of places where we're using static inputs for attributes like top and height. We can start transitioning these to MediaQuery where relevant to maintain a dynamic UI. I also left a comment about reducing the nesting of Container widgets, which should help make the code cleaner.

PR #239, opened 2024-12-09
Added functionality of blocking/deleting users

Shengle-Dai:
### Summary 

This PR allows administrators to delete/ban users on the admin page.
- Added the new button svgs and fully functioning buttons to the HTML
- Added banUser socket to ServerData.tsx

### Test Plan 

Tested on local device:
Clipboard Dec 8, 2024 at 10 16 PM

### DIFF
--- a/admin/src/components/Users.tsx
+++ b/admin/src/components/Users.tsx
@@ -49,6 +49,41 @@ const SAVE_SVG = (
   </svg>
 );
 
+const DELETE_SVG = (
+  <svg
+    height="20"
+    viewBox="0 0 20 20"
+    width="20"
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <path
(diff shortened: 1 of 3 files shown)


admin/src/components/Users.tsx
```diff
@@ -163,6 +200,50 @@ function getColumns(setRowsData: any, serverData: any) {
...
+          <button
+            title="Delete"
+            style={styles.cancelButton}
+            onClick={(e) => {
+              e.stopPropagation();
+
+              if (
+                window.confirm("Are you sure you want to delete this user?")
+              ) {
+                setRowsData((prevRows: any[]) =>
```
shah-esha, line 237: Amazing job implementing the blocking feature! One thing I would consider is using a map for the row data (such as prevRows) instead of an array to optimize updates for larger datasets. This could allow for constant-time lookups and updates.

PR #245, opened 2025-02-14
Document AuthFlow and Glitchy Login Fix 

3TTemi:
### Summary 

PR to fix high priroity bug: 

- [x] Implemented fix that clears navigator stack when user disconnects to offset page overlaps
- [x] Added Documentation for login/authetnications Flow  

Remaining TODOs:

- [ ] Enact the test plan below 

### Test Plan 

Do the following test on test flight: 

Specifically by going into a challenge and then swiping out to another app, waiting for a few minutes for a disconnect, and then returning back to the CornellGo app. 

### Notes 

During testing, I found there there may be another navigation bug related to overlaying pages which occurs after the user signs out and logs back into the app. It is visible through a back button on the search filter home page that should not be there. Further attention needs to be looked into this.

### DIFF
--- a/game/lib/gameplay/gameplay_map.dart
+++ b/game/lib/gameplay/gameplay_map.dart
@@ -23,24 +23,28 @@ import 'package:game/model/group_model.dart';
 import 'package:game/model/event_model.dart';
 import 'package:game/model/challenge_model.dart';
 
-/// GameplayMap Widget
-///
-/// Displays an interactive map for gameplay, showing the user's location,
-/// hint circles, and challenge-related information.
-///
-/// @remarks
-/// This widget is responsible for handling user location updates, displaying
-/// hints, and managing the challenge completion process. It uses Google Maps
-/// for rendering the map and integrates with various game-related models and
-/// APIs to provide a seamless gameplay experience.
-///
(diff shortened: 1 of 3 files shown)


ambers7 (commented):
The documentation is very clear and detailed. The examples help clarify the use of code as well.

game/lib/gameplay/gameplay_map.dart
```diff
@@ -358,13 +362,16 @@ class _GameplayMapState extends State<GameplayMap> {
...
                   if (client.serverApi == null) {
                     WidgetsBinding.instance.addPostFrameCallback((_) {
-                      Navigator.pushReplacement(
-                          context,
-                          MaterialPageRoute(
-                              builder: (context) => SplashPageWidget()));
-                      displayToast("Lost connection!", Status.success);
+                      // Clear entire navigation stack and push to login screen
+                      Navigator.of(context).pushAndRemoveUntil(
+                        MaterialPageRoute(
```
shah-esha, line 370: Great job on fixing the overlapping pages issue! This should do the trick. However, I believe the app might still navigate back to the login page when put in the background for a few seconds. We might need to look into a solution for that (maybe using token-based authentication to persist login sessions properly). Overall, great work!

PR #246, opened 2025-02-15
Update gradle and fix bugs inside game/android

ambers7:
### Summary 
PR to update gradle and flutter for Android. 

- [x] change deprecated Flutter Gradle plugins from imperative apply script method to declarative plugins (https://docs.flutter.dev/release/breaking-changes/flutter-gradle-plugin-apply#androidsettings-gradle)
- [x] fix namespace not found error  
- [x] remove platform_device_id and use device_info_plus package instead

Remaining TODOs:

- [ ] resolve kotlin 17 and java 1.8 version compatability error

### Test Plan 
- get gradle to build successfully on windows laptop 

### Notes 
- had a java 21 vs java 17 error and was able to fix by temporarily changing the env:JAVA_HOME variable, might need to do that every time
```$env:JAVA_HOME= (path to java installation)```
```$env:PATH="$env:JAVA_HOME\bin;$env:PATH" ```

### DIFF
--- a/game/android/app/build.gradle
+++ b/game/android/app/build.gradle
@@ -39,8 +39,11 @@ android {
     }
 
     defaultConfig {
-        minSdkVersion 22
-        targetSdkVersion 34
+        minSdk 22
+        targetSdk 34
+//
+//        minSdkVersion 22
+//        targetSdkVersion 34
         versionCode flutterVersionCode.toInteger()
         versionName flutterVersionName
     }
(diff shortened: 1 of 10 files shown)


3TTemi:
Changes look good! Hopefully, they can permanently fix some of these compatibility issues.

For the test plan, I would definitely add that we should try building the project on both Windows and Mac, as well as testing on iPhone and Android simulators. I know that changing these files can sometimes break things.

ambers7:
Fixed kotlin 17 and java 1.8 compatability error as well as the signing configuration error with appminifiy.

PR #248, opened 2025-02-21
Glitchy Login Fix

shah-esha:
### Summary 

This PR is a fix for a high priority bug

- [x] Enabled reconnection to ensure the app can recover gracefully from disconnects whenever possible

Remaining TODOs:

- [ ] Testing

### Test Plan 

Test using TestFlight:

Swipe out of the app so it's in the background. Wait a few minutes and enter the app again. The bug is fixed if the app doesn't divert back to the login page.

### Notes 

We want to keep an eye out to make sure these changes don't break other parts of the app.


3TTemi (approved):
Changes look good to go as we reviewed during dev sesh and ready for testing

PR #251, opened 2025-03-20
Fix Update Username Bug 

ambers7:
### Summary 

This PR fixes the username not being able to change. 

- lets the update button be pressed after the username is changed
- checks that new username is unique
- if the username is not unique, displays an error message below the username text field saying it must be unique

Remaining TODOs:
- [ ] keep testing bug fix
- [ ] in notes

### Test Plan 
- test update button is enabled after username field is edited
- change username to one that does not exist and update it
- confirm the new updated username is used everywhere in the app instead of the old name
- ensure the update button is not enabled when a non-unique username is inputted and the error message is displayed

## Notes: 
We discovered that the bug was exclusive to TestFlight (works in simulator but not in TestFlight), so figure out why/how


3TTemi:
Nice work integrating both the frontend and backend changes to fix this! That said, while reviewing the code, I found the handling of null values in the edit_profile page a bit confusing in terms of when it returns true or false. I’d recommend revisiting the comments to make that logic clearer.

Also, make sure to review your already written comments — we don’t need to reference what happened "previously," since anyone looking at the codebase will only care about its current state.

PR #254, opened 2025-03-22
Update User Location Faster 

3TTemi:
### Summary 

This pull request aims to decrease the time in which it takes for a the user location to load when entering a challenge 

- [x] Locally store and access devices last known location (uses this rather than just mock/general location previosuly being used) 

### Test Plan 

Test on TestFlight implementation of app under differnet conditions:

- Right after powering on/restarting the phone 
- After opening an app that uses location services (ex: Google Maps) 
- In nomral use conditions


shah-esha (approved):
Looks good!

shah-esha (approved):
Merged master w/ this branch.

PR #255, opened 2025-03-24
Fix achievement bugs and add in all majors/colleges

shah-esha:
### Summary 

This pull request aims to help prepare CGO for launch by fixing an achievement bug and adding in all Cornell majors and colleges. The achievement icon was not changing to gold after being successfully completed, which I implemented a quick fix for. The "Major" dropdown would also be null for colleges that did have majors, so I fixed that as well.

- [x] Achievements display the correct icon depending on progress
- [x] Added in all majors/colleges in both registration and edit-profile pages
- [x] Dropdown works for colleges w/ majors
- [x] Tested changes on simulator  

Remaining TODOs:

- [ ] Test on testflight


3TTemi (dismissed):
Changes look good to merge! Will try and do merge conflicts with master

3TTemi (dismissed):
Merged changes with master, good to go.

PR #256, opened 2025-04-08
Fixing Potential Dependency Issues

shah-esha:
### Summary 

This pull request is to try fixing the Testflight Issue we are having, where there is an error authenticating users/signing new users up. I updated main.dart so that it uses Flutter Config Plus instead of Flutter Config and fixed other issues that were causing errors in simulation. We still need to test these changes in Testflight to see if that fixes the issue we're facing.

- [x] updated imports
- [x] fixed duplicates in pubspec.lock

Remaining TODOs:

- [ ] at least one other person tests locally
- [ ] merge into master and update Testflight build to see if there was improvement

### Notes 

If we are still having issues with signing in, we may need to pause and revert to an older version of the app and build from there for the sake of soft launch.

### DIFF
--- a/game/ios/Podfile
+++ b/game/ios/Podfile
@@ -1,5 +1,5 @@
 # Uncomment this line to define a global platform for your project
-platform :ios, '14.0'
+platform :ios, '15.2'
 
 # CocoaPods analytics sends network stats synchronously affecting flutter build latency.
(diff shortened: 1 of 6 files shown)


3TTemi (approved):
Looks good, made some small changes with duplicate code.

PR #257, opened 2025-04-14
Edit Profile Page Updates

3TTemi:
### Summary 

[x] API Configuration Enhancements
- Improved environment variable handling with proper API URL initialization
- Fixed initialization order for Flutter bindings and plugins
- Initialized API client with late variables for better memory management

[x] Edit Profile UI Improvements
- Fixed overflow issues by implementing SingleScrollView for the profile form
- Made the Update button responsive with full width layout
- Optimized button UI by removing unnecessary nested layouts

[X] Fix logic for being able to update username (was previously grayed out even when field changed) 

[x] Added Documetnation to Edit Profile Page 

### Test Plan 

API Configuration:
- Tested app startup with and without .env file
- Verified correct API URL is used in different environments
- Confirmed app properly handles API client initialization

Edit Profile UI:

- Verified form is fully scrollable on different device sizes (iPhone SE and Iphone 15 pro)
- Verified update functionality works correctly after UI changes

### DIFF
--- a/game/lib/profile/edit_profile.dart
+++ b/game/lib/profile/edit_profile.dart
@@ -6,6 +6,16 @@ import 'package:game/progress_indicators/circular_progress_indicator.dart';
 import 'package:game/model/user_model.dart';
 import 'package:provider/provider.dart';
 
+/**
+ * `EditProfileWidget` - A form interface for editing user profile information.
+ * 
+ * @remarks
+ * This widget provides a complete form for users to edit their profile details including
+ * username, college, major, and graduation year. It uses dropdown selectors for predefined
+ * options and text input for the username. The widget automatically detects changes to
+ * enable/disable the update button and communicates with the UserModel to persist changes.
+ * 
+ */
 class EditProfileWidget extends StatefulWidget {
   EditProfileWidget({
     Key? key,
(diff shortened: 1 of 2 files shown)


game/lib/profile/edit_profile.dart
```diff
@@ -160,18 +159,28 @@ class _EditProfileState extends State<EditProfileWidget> {
           String? currCollege = userModel.userData?.college;
           String? currMajor = userModel.userData?.major;
 
-          newUsername = currUsername;
-          newYear = currYear;
-          if (newYear != null && newYear!.isEmpty) {
-            newYear = null;
+          // Initialize fields only if they haven't been already
```
shah-esha, line 186: I think this is a good idea to try tackling the changing username issue we're running into right now. If newUsername is being reinitialized to currUsername every time there's a change, it makes sense why the app isn't able to detect that update. However, it might be a good idea to think about why the code was working for the dropdowns even before we changed this code.

Also, I think the code for newYear and newCollege can be cleaned up a little:

`if (newCollege == null) {
  newCollege = (currCollege?.isNotEmpty ?? false) ? currCollege : null;
}`

Overall, awesome job Temi!

PR #258, opened 2025-04-14
Timer feature

ambers7:
### Summary 
This pull request is the first step towards implementing the timer feature for challenges. 

- [x] implemented timer that counts down from 1:00 inside the map page 

Remaining TODOs:

- [ ] have the circle indicate timer progress as well 
- [ ] show a warning when 5 minutes are left 
- [ ] change the timer depending on the challenge
- [ ] have challenges with timers and without timers
- [ ] show the challenge over page once the timer runs out and have an option to add 5 minutes

### Test Plan 
- create challenges with timers under 5 minutes, timers over 5 minutes, and no timer
- start each challenge - for each, make sure that the timer functions correctly and that the warning screen shows up when the timer hits 5 minutes. After the challenge is over, if the user decides to add 5 more minutes, make sure that their points are deducted.


game/lib/gameplay/gameplay_map.dart
```diff
@@ -479,13 +479,16 @@ class _GameplayMapState extends State<GameplayMap> {
...
                 alignment: Alignment.center,
                 children: [
-                  SvgPicture.asset(
-                    "assets/icons/timerbg.svg"
-                  ),
+                  (currentTime < 300) ?
+                  SvgPicture.asset( //turn timer red when less than 5 minutes left
+                    "assets/icons/timerbg_red.svg"
+                  ) :
+                    SvgPicture.asset( "assets/icons/timerbg.svg"
```
shah-esha, line 486: Awesome job Amber! I think one suggestion I have here is that instead of using static SVGs for the timer background, consider switching to a responsive Container with dynamic colors. This will scale better, is less taxing on memory/processing, avoids the need for external assets, and might make more sense intuitively. Otherwise, awesome job and progress on the timer feature!

PR #267, opened 2025-09-21
Add 2030 graduation year & Alumni option

jiminkim214:
### Summary 

- [x] added options for user to select graduation year 2030 (for architecture students) and an alumni option during registration and when editing their profile

Depends on #{number of PR}

### Test Plan 
- [x] test on different simulators

### DIFF
--- a/game/lib/details_page/details_page.dart
+++ b/game/lib/details_page/details_page.dart
@@ -68,7 +68,7 @@ class _DetailsPageWidgetState extends State<DetailsPageWidget> {
     "Weill Cornell Medicine"
   ];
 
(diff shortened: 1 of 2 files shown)


ambers7:
Good job Jimin! One thing to consider for the future is how we might change how we choose the graduation years users can select, so that we won't have to keep updating the list every year. We also might want to have the "Alumni" option as a separate checkbox, so Alumni don't have to scroll to the bottom of the year list.

3TTemi (approved):
Changes look good and work in the app, other comments are minor issues outisde of scope of issue.

PR #268, opened 2025-09-21
Minimum Character Requirement for Usernames

ambers7:
### Summary 

Required users to have a username of at least three characters in the following scenarios:
- [x] when first creating their username during onboarding
- [x] when editing their profile 

### Test Plan 
- have users try to create accounts and edit their profiles with usernames of length 0, 1, 2, and 3 characters
- try the above with multiple emulators/simulators

### DIFF
--- a/game/lib/details_page/details_page.dart
+++ b/game/lib/details_page/details_page.dart
@@ -38,7 +38,10 @@ Key features include:
 - Form validation using [_formKey]
 - State-managed inputs with real-time updates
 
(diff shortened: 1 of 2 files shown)


3TTemi:
Tested changes in the simulator and they look good! Good job also integrating your comments into the pre-existing ones. Slight typo was made in a edit_profile comment should be "been" instead of "bene".

3TTemi (approved):
After changes to previous comment it looks good!

PR #272, opened 2025-10-03
Backend for Game Onboarding

ambers7:
### Summary 

Created the backend for storing, reading, and updating a user's onboarding status (if they have completed it or not). 
This pull request is the first step towards implementing user onboarding when they first download CornellGO. 

### Test Plan 
- Manual database testing (done): run backend locally (npm run start:dev) 
- [ ] When onboarding frontend is done check the values of hasCompletedOnboarding for when they finish the onboarding, and when they decide to redo the onboarding. 

### Breaking Changes  
- Database schema change (added hasCompletedOnboarding to User)


jiminkim214 (commented):
Great work! The onboarding status logic and DTO additions are cleanly integrated. I'm wondering if there any potential migration concerns for existing users regarding the new hasCompletedOnboarding field?

server/prisma/migrations/20250922023823_add_user_fields/migration.sql
3TTemi, line 1: Was confused on what was happening here with keys and constraints. If i'm correct Prisma generated these changes not you, if so you should list that somewhere here so that it is clear.

3TTemi:
Good job on this PR. Reviewing the logic, I believe this will successfully add the ability to track and reset a user's onboarding status.

However, I noticed that only about 8 files had actual changes, and the rest were very small, like adding commas. Do you know why this happened? Are you using some type of formatter that is not Prettier (the one we adhere to in the codebase)?

I also left a comment about differentiating between developer-made and generated code to ensure clarity for other developers. Otherwise, the logic looks good!

3TTemi:
The reasoning why server code style test was failing even after running `npm run formatall`/`dart format lib `was because of a Prettier version mismatch.
 
Local: Prettier 2.8.8 (global)
Docker: Prettier 3.6.2 (from package.json)
Prettier 3.6.2 has different formatting rules, so it flagged files that 2.8.8 thought were fine

After I installed dependencies with npm install --legacy-peer-deps to get Prettier 3.6.2 locally and then reran formatting with Prettier 3.6.2 this was fixed

PR #273, opened 2025-10-04
Jk2756/longer descp

jiminkim214:
### Summary 

- [x] updated database schema to include long description field
- [x] updated backend DTOs and services
- [x] updated backend gateway
- [x] regenerated frontend DTO
- [x] updated frontend UI components 

Remaining TODOs:

- [ ] add tests for new description handling
- [ ] verify api alignment between backend and frontend

Depends on #{number of PR}

### Test Plan 
- [x] tested through mock data and simulator
- [x] confirmed correct rendering of both short and long descriptions in UI

### Blockers 

### Breaking Changes  

- Database schema change (new longDescription field)


ambers7 (commented):
Good job Jimin! I liked your documentation for the DTOs in event.dto.ts; it might also be helpful to add something in to event.service.ts about what differs a description from a long description (ex. character limit) and if there's a limit to the length of the long description.

3TTemi:
@jiminkim214 
Now that we have learned testing make sure to add some tests to verify that long description is working correctly. Currently our event service is under tested so it would be good to make some basic tests for upsertEventFromDto, getEventById, and removeEvent from eventService which you can use to indirectly test long description.

3TTemi (approved):
Tested changes in local dev server, approving to push to production.

PR #304, opened 2026-03-30
Notification Reminder for Events

jjasonguo:
### Summary 

Notications for event rsvp with e2e testing. Note the schema migration includes something from a previous migration but cursor was telling me to keep it cuz its a drift and needed to be added in eventually anyway.

### Test Plan 

e2e tests

### Blockers 

### DIFF
--- a/server/src/campus-event/campus-event.e2e-spec.ts
+++ b/server/src/campus-event/campus-event.e2e-spec.ts
@@ -324,7 +324,7 @@ describe('CampusEventModule E2E', () => {
 
       sendToUserMock.mockClear();
       sendToUserMock.mockImplementation(
-        () => new Promise((resolve) => setTimeout(() => resolve(true), 100)),
+        () => new Promise(resolve => setTimeout(() => resolve(true), 100)),
       );
 
       await Promise.all([
--- a/server/src/campus-event/campus-event.e2e-spec.ts
+++ b/server/src/campus-event/campus-event.e2e-spec.ts
@@ -296,6 +296,42 @@ describe('CampusEventModule E2E', () => {
       expect(sendToUserMock).not.toHaveBeenCalled();
     });
 
+    it('prevents duplicate sends during concurrent cron runs', async () => {
+      if (!campusEventTableExists || !testUserId) return;
+
+      const startTime = new Date(Date.now() + LEAD_TIME_MS + 90_000);
+      const endTime = new Date(startTime.getTime() + 3600_000);
+      const ev = await campusEventService.createEvent({
+        title: 'Concurrent Reminder Event',
+        description: 'Should only send one reminder',
+        startTime: startTime.toISOString(),
+        endTime: endTime.toISOString(),
+        locationName: 'Statler Hall',
+        latitude: 42.4458,
+        longitude: -76.4821,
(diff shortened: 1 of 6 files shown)


server/src/campus-event/rsvp-reminder.service.ts
```diff
@@ -0,0 +1,65 @@
...
+    this.logger.log(`Sending reminders for ${rsvps.length} RSVPs`);
+
+    for (const rsvp of rsvps) {
+      const { title, startTime } = rsvp.campusEvent;
+      const timeStr = startTime.toLocaleTimeString([], {
+        hour: '2-digit',
+        minute: '2-digit',
+      });
+
+      await this.notificationService.sendToUser(
```
3TTemi: If any of the loops .sendToUser or the below .update fail then the entire loop aborts, instead we should use try catch inside for loop.

server/src/campus-event/rsvp-reminder.service.ts
```diff
@@ -0,0 +1,65 @@
...
+      },
+    });
+
+    if (rsvps.length === 0) return;
+
+    this.logger.log(`Sending reminders for ${rsvps.length} RSVPs`);
+
+    for (const rsvp of rsvps) {
+      const { title, startTime } = rsvp.campusEvent;
+      const timeStr = startTime.toLocaleTimeString([], {
```
3TTemi: LocalTime on the server might not be EST.

To make sure our output is always EST add these:
'en-US' which locks the format to 12-hour with AM/PM
timeZone: America/New_York' which ensures it always shows Eastern Time

server/src/campus-event/campus-event.e2e-spec.ts
```diff
@@ -220,6 +222,116 @@ describe('CampusEventModule E2E', () => {
...
+          .deleteMany({ where: { campusEventId: reminderEventId } })
+          .catch(() => {});
+        await campusEventService.deleteEvent(reminderEventId).catch(() => {});
+      }
+    });
+
+    it('sends reminder for RSVP within the cron window', async () => {
+      if (!campusEventTableExists || !testUserId) return;
+
+      const startTime = new Date(Date.now() + LEAD_TIME_MS + 60_000);
```
3TTemi: Thinking about it now we should add some small buffer time to start and endTime in case one of the server cron pings are late for example. Will cause window overlaps but should be fine because we have reminderSent flag that prevents duplicates.

server/src/campus-event/rsvp-reminder.service.ts
```diff
@@ -0,0 +1,65 @@
...
+      });
+
+      await this.notificationService.sendToUser(
+        rsvp.userId,
+        `Upcoming Event: ${title}`,
+        `Starts at ${timeStr} — see you there!`,
+        { campusEventId: rsvp.campusEventId },
+      );
+
+      await this.prisma.eventRSVP.update({
```
gselina: Great work Jason! I think there might be a potential race condition where users could get duplicate notifications if an rsvp is processed multiple times before it’s marked as sent. Otherwise the implementation looks good!

3TTemi (changes requested):

3TTemi (dismissed):

PR #306, opened 2026-04-12
New Home Page Overlays

gselina:
### Summary 

This pull request is the first step towards implementing the new home page by building the frontend overlay UI for the new Home page. Changes are UI-only: created the components and layout, but real functionality has not been implemented (API calls, no filtering logic, navigation behavior)

- [x] implemented search bar with filter button
- [x] created static filter chips as reusable chip components (for Food, Speakers, Concerts, etc)
- [x] created reusable event pins for each state
- [x] home tab shows map + overlays

Remaining TODOs:

- [x] update selected pin icon (currently all burger)

### Test Plan 

https://github.com/user-attachments/assets/8496aad4-a831-49c9-8318-1fdd7f1bb611


jjasonguo:
Looks very good Selina. Very clean changes when testing on my emulator. We should consider what the journey tap should lead to next

3TTemi (changes requested):
.

game/lib/navigation_page/bottom_navbar.dart
```diff
@@ -64,7 +64,12 @@ class _BottomNavBarState extends State<BottomNavBar> {
     fontSize: 30,
     fontWeight: FontWeight.bold,
   );
-  late final List<Widget> _widgetOptions;
+  static List<Widget> _widgetOptions = <Widget>[
```
3TTemi: Instead of this static list, lets. keep to the same format as  _widgetOptions is in master (with `late final` and being in `initstate()`)

This way parameters can be used which is needed for another pages functionality

game/lib/navigation_page/home_map_page.dart
```diff
@@ -0,0 +1,1172 @@
...
+  final targetW = selected ? 60.0 : 50.0;
+  final targetH = selected ? 71.0 : 53.0;
+  final sw = targetW / refW;
+  final sh = targetH / refH;
+  final refScale = selected ? (sw > sh ? sw : sh) : (sw < sh ? sw : sh);
+  final s = refScale * scaleMultiplier;
+  return ui.Size(vb.width * s, vb.height * s);
+}
+
+const _kExamplePins = <_ExampleMapPin>[
```
3TTemi, line 140: Nice use of examples so this can be demoed, just add a TODO comment so we know we need to come back to this when we need to integrate with real event data

3TTemi:
Great work! I tested this in the simulator, and the UI looks really good.

To align with our coding standards and make this easier to review/maintain, you should make these improvements though:

- Refactor home_page.dart: the file is really large. You can break it up into smaller, modular Dart files like: Category Chips, Map Pin Utilities, and Search Bar
- Documentation: Add comments or headers to define the different sections of the code. This will help guide future contributors through the logic more easily

3TTemi (changes requested):

PR #307, opened 2026-04-13
Jhg294/events bottom sheet

jjasonguo:
### Summary 

This pull request is the first step towards implementing the new event based ui of cornellgo. It creates the bottom sheet page that will display event. Currently, the GO button has no functionality. Retrieved events are sorted by end time first, then location. subject to change in the future. They have a hosted by with default black circle for now too.

Update 4/30:

This PR now has the events home page incorporated, with working go functionality and gcal integration for future events. Things to consider are when we want the event to say "GO" vs "add to gcal", and also whether the dotted line for navigation should be straight or like an actual path using google maps api.

Remaining TODOs:


ambers7:
Good job Jason! I would elaborate more on the TODO, like including the "Set Reminders feature", the dotted line showing the path to the event, and adding in more icons for the events besides the restaurant one (once they are made), and adding in pins for the events where they are located (maybe just a placeholder one on the screen until the backend is connected). 
One thing to note is that ```apiClient.serverApi?.setCurrentEvent(SetCurrentEventDto(eventId: ""));``` is inside ```_filteredEvents```, which gets called many times as Flutter runs ```build```. It would be better to make this a separate function and call it outside of ```build``` instead.

3TTemi (dismissed):
.

3TTemi:
UI looking good!

But first thing I noticed when trying out in simulator is that the touch area for the bottom sheet is buggy. Seem to only be able to move it from the list view not the header or drag bar

game/lib/navigation_page/events_draggable_sheet.dart
```diff
@@ -0,0 +1,506 @@
...
+      if (diff.inMinutes > 0) return 'in ${diff.inMinutes} min';
+      return 'soon';
+    } catch (_) {
+      return '';
+    }
+  }
+
+  List<EventDto> _filteredEvents(
+    UserModel userModel,
+    EventModel eventModel,
```
3TTemi, line 108: Lowkey not your fault since it wasn't developed at the time you made this branch but were supposed to be using campuseventModel and using the campuseventDTO as well not eventModel. This makes it so its shows actual events and not challenges.

If you merge master you will get these models and DTOs

(need to change eventModel name so this is more clear, will do)

3TTemi (changes requested):

jjasonguo:
thanks for feedback, addressed draggable area and repeating api call in the build. will continue with the correct campuseventmodel when merging changes with Selina's!

PR #310, opened 2026-04-13
Rename ><.png to squinty_eyes.png for Windows

ambers7:
Rename ><.png in buildabear since that file name is invalid on Windows computers


3TTemi (approved):
Might be the smallest pr commit on CGO codebase 😂 but clearly needed, LGTM

PR #319, opened 2026-04-27
Popup UI - “Dont Miss Out”

gselina:
### Summary 

This pull request implements a reusable pop-up widget that displays an envelope-style notification with configurable text

- [x] implemented widget that accepts three dynamic parameters:
    - Title text (shown at the top of the popup)
    - Body text (shown inside the envelope graphic)
    - Button label text
- [x] when the label is “Turn On Notifications”, button opens the app’s settings page on the user’s device

### Test Plan 

https://github.com/user-attachments/assets/6743d7b9-5f68-4691-8809-a2c65f9d74d6


game/lib/navigation_page/bottom_navbar.dart
```diff
@@ -80,6 +82,19 @@ class _BottomNavBarState extends State<BottomNavBar> {
       GlobalLeaderboardWidget(),
       ProfilePage(),
     ];
+
+    // TODO: testing purposes only, remove later
+    if (kDebugMode) {
+      WidgetsBinding.instance.addPostFrameCallback((_) async {
+        if (!mounted) return;
```
jjasonguo, line 89: Looks great Selina. Last thing would be to determine the frequency we want to show this pop-up, maybe if they don't want notifications we shouldn't bother them every single time they open the app. Additionally, when notifications ARE turned on, the popup still appears. There should be a way to see if they turned them on already.

jjasonguo (changes requested):
Looks good check my one comment

ambers7 (commented):
Good job Selina! Make sure that you're using colors from `game/lib/constants/colors.dart` instead of hardcoding them. It might also be good to add a quick description of what CTA means in the documentation.

PR #320, opened 2026-04-27
Admin Dashboard Campus Event Creation

ambers7:
### Summary 

This pull request is implements creating campus events in the admin dashboard. Admins can create or edit an event by filling out a form with information, approve/reject an event, and delete an event. 

- [x] added "Campus Event' to the sidebar which leads to the Campus Event page
- [x] create a form with required fields `title, description, start time, end time, location name, latitude/longitude (via MapEntryForm), categories ` and optional fields `image URL, all-day toggle (default false), address, check-in radius (default 100m), tags, organizer name, organizer email, check-in method dropdown, points for attendance (default 10), featured toggle (default false), registration URL`
- [x] once the form is filled out, the campus event is created in the backend with the given information
- [x] admins can approve/reject/archive events. If an event is rejected, a rejection reason must be filled out. 
- [x] wired a `onUpdateCampusEventData` listener so an update to one event updates the list without re-requesting the whole list 

Remaining TODOs:

- [x] check that rejection reasons are required when rejecting an event

### DIFF
--- a/admin/package-lock.json
+++ b/admin/package-lock.json
@@ -84,7 +84,6 @@
       "version": "7.21.3",
       "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.21.3.tgz",
       "integrity": "sha512-qIJONzoa/qiHghnm0l1n4i/6IIziDpzqc36FBs4pzMhDUraHqponwJLiAKm1hGLP3OSB/TVNz6rMwVGpwxxySw==",
-      "peer": true,
       "dependencies": {
         "@ampproject/remapping": "^2.2.0",
         "@babel/code-frame": "^7.18.6",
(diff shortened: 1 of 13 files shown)


gselina:
Great work Amber! One thing I noticed is that the use of hardcoded form indices could potentially break validation if the fields were to change in the future. But otherwise the implementation looks good!

jiminkim214:
> Great work Amber! One thing I noticed is that the use of hardcoded form indices could potentially break validation if the fields were to change in the future. But otherwise the implementation looks good!

I agree with this. The logic around form[18] and form[19] works right now, but it feels a bit fragile since adding, removing, or reordering fields could silently break the approval/rejection behavior. It might be worth using helper accessors so it is less dependent on array position. Other than that, the overall flow looks solid!

